PR #37, opened 2016-10-26
Sets up exponential backoff for event dispatch and datafile download

jophde:
Datafile download and event dispatch were both happening synchronously on background threads.  We simply need to make the client's request method handle exponential backoff internally.  As much as possible was moved into the shared library since the event client and datafile client both use the same client core.

Tested e2e manually via logs and putting a phone into airplane mode.  There is no easy way to programmatically cause network failures.  Added more unit test coverage.

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
@@ -39,38 +39,50 @@ class DataFileClient {
         this.logger = logger;
     }
 
-    String request(String urlString) {
-        HttpURLConnection urlConnection = null;
-        try {
-            URL url = new URL(urlString);
-            logger.info("Requesting data file from {}", url);
-            urlConnection = client.openConnection(url);
+    String request(final String urlString) {
+        Client.Request<String> request = new Client.Request<String>() {
+            @Override
+            public String execute() {
+                HttpURLConnection urlConnection = null;
+                try {
+                    URL url = new URL(urlString);
+                    logger.info("Requesting data file from {}", url);
+                    urlConnection = client.openConnection(url);
 
-            client.setIfModifiedSince(urlConnection);
+                    client.setIfModifiedSince(urlConnection);
 
-            urlConnection.setConnectTimeout(5 * 1000);
-            urlConnection.connect();
+                    urlConnection.setConnectTimeout(5 * 1000);
+                    urlConnection.connect();
 
-            int status = urlConnection.getResponseCode();
-            if (status >= 200 && status < 300) {
-                client.saveLastModified(urlConnection);
-                return client.readStream(urlConnection);
-            } else if (status == 304) {
-                logger.info("Data file has not been modified on the cdn");
-                return null;
-            } else {
-                logger.error("Unexpected response from data file cdn, status: {}", status);
-                return null;
+                    int status = urlConnection.getResponseCode();
+                    if (status >= 200 && status < 300) {
+                        client.saveLastModified(urlConnection);
+                        return client.readStream(urlConnection);
+                    } else if (status == 304) {
+                        logger.info("Data file has not been modified on the cdn");
+                        return "";
+                    } else {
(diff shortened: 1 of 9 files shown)


android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
```diff
@@ -39,38 +39,50 @@
...
-            if (urlConnection != null) {
-                urlConnection.disconnect();
-            }
-        }
-    }
+        };
 
+        // If the response was a 304 Not Modified an empty string is returned
+        // Consumers of this method expect either a valid datafile or null
+        String response = client.execute(request, 2, 3);
```
mikeng13: what are those two numerical params you are passing in? Can you make them into constants or leave some comments about them?

android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
```diff
@@ -39,38 +39,50 @@
...
+        if (response != null && response.isEmpty()) {
+            response = null;
```
mikeng13: nit: we can just `return null` here

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventClient.java
```diff
@@ -41,29 +41,40 @@
...
+                    logger.error("Unable to send event: {}", event, e);
+                    return Boolean.FALSE;
+                }
             }
-        } catch (IOException e) {
-            logger.error("Unable to send event: {}", event, e);
-            return false;
+        };
+
+        Boolean success = client.execute(request, 2, 5);
```
mikeng13: same comment here about the magic numbers

test-app/src/main/java/com/optimizely/ab/android/test_app/MyApplication.java
```diff
@@ -66,7 +66,7 @@ public void onCreate() {
         // must match the project id of the compiled in Optimizely data file in rest/raw/data_file.json.
         optimizelyManager = OptimizelyManager.builder(PROJECT_ID)
                 .withEventHandlerDispatchInterval(3, TimeUnit.SECONDS)
-                .withDataFileDownloadInterval(30, TimeUnit.MINUTES)
+                .withDataFileDownloadInterval(30, TimeUnit.SECONDS)
```
mikeng13: lol good fix

jophde: This is just in the test app.  Users can set this to whatever they want.  Faster is better for our testing.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
jophde: The first number is the base.  The second number is the power. This is used to calculate the timeout for the backoffs.  Each client could configure a different base and power.  For example the datafile client uses 2^3.  Event dispatch uses 2^5.  In the later the last backoff try, 5, will have a 32 second timeout.  The max backoff for 2^3 is 8 seconds.

jophde: I can pull them out as constants in the DataFileClient and the EventClient and pass the constants though.  However, they are only used in one place.

mikeng13: Ah ok you don't have to pull out as constants but just put a note in the doc as to what those numbers are

android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
mikeng13: I meant returning `null` inside of the `if` clause as opposed to setting the `response` to null and then letting it return later. Just personal preference though, you can ignore this.

jophde: Oh ok I see what you mean.

PR #44, opened 2016-11-22
Refactor initialization API to conform to shared design with iOS SDK.

mikeng13:
@optimizely/fullstack-devs 

Refactored the initialization API to match the API design we agreed upon for both iOS and Android SDKs.


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -118,6 +111,63 @@ void setOptimizelyStartListener(@Nullable OptimizelyStartListener optimizelyStar
     }
 
     /**
+     * Initialize Optimizely Synchronously
```
alda-optimizely: iOS also has a third API with no parameters. This initializes an Optimizely object with the local datafile. If no local datafile exists, than a dummy Optimizely object is returned.

https://docs.google.com/document/d/1ntpMeUbNgYXSq-fEHyATK1iknhutHpsXUBuU4uIOPF8/edit

We may also need a helper method or property called: isDatafileCached

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -118,6 +118,63 @@ void setOptimizelyStartListener(@Nullable OptimizelyStartListener optimizelyStar
...
+     * <p>
+     * Instantiates and returns an {@link OptimizelyClient} instance. Will also cache the instance
+     * for future lookups via getClient
+     * @param activity an Activity, used to automatically unbind {@link DataFileService}
+     * @param datafile the datafile
+     * @return an {@link OptimizelyClient} instance
+     */
+    public OptimizelyClient initialize(@NonNull Activity activity, @NonNull String datafile) {
+        if (!isAndroidVersionSupported()) {
```
vraja2: should we log here?

mikeng13: We already log inside of that method :)

alda-optimizely:
i'd differ to @jophde for the final ok though.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
```diff
@@ -89,6 +89,14 @@
     }
 
     /**
+     * Check that this is a valid instance
```
jophde: Wouldn't optimizelyClient always be non null since it's always created by default with a null instance of optimizely core?

jophde:
LGTM

PR #46, opened 2016-11-25
Nulls out the dataFileService connection

jophde:
Since we are using the nullability of the dataFileServiceConnection to control whether we can bind we need to null it out so that subsequent start and initialize calls still try to bind the service.  

Generally, most apps will probably have one OptimizelyManager instance.  Without this change you couldn't call initialize on more than one activity.


android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/DataFileReschedulerTest.java
```diff
@@ -24,6 +24,7 @@
 import com.optimizely.ab.android.shared.Cache;
 
 import org.junit.Before;
+import org.junit.Ignore;
```
mikeng13: nitpick: this still needed?

PR #47, opened 2016-11-28
Use old constructor for db manager

jophde:
Ebay is reporting crash on ICS devices because we are using a certain constructor.  We are only passing null so just use the old version that is compatible with all devices. 

"On devices running Android 4.0.x, we got a NullPointerException, the EventSQLiteOpenHelper is being created.
In ICS the default parameter for the DatabaseErrorHandler was not null but new DefaultDatabaseErrorHandler() (Source ICS vs. Source N). Since you are calling not the default overloaded constructor, a NullPointerException is thrown. You should instead just use the constructor without the explicit DatabaseErrorHandler. See the attached stacktrace for more details."


jophde:
I reproed the crash on a local emulator running 14.  Removing the null param from the constructor call does indeed fix this.  Ran it successfully on the emulator after this change.

.idea/misc.xml
```diff
@@ -37,7 +37,7 @@
     <ConfirmationsSetting value="0" id="Add" />
     <ConfirmationsSetting value="0" id="Remove" />
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_7" default="true" assert-keyword="true" jdk-15="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
mikeng13: What is this property for?

jophde: This should be git ignored.  Will remove.  It seems to always come back though.

PR #51, opened 2016-12-09
Add live variables

vraja2:
@optimizely/fullstack-devs


jophde:
Just make sure it launches on API 14

mikeng13:
@jophde @vraja2 I checked on API 14. We're good to go.

PR #53, opened 2016-12-15
Make getDatafileUrl static for easier access.

mikeng13:
@optimizely/fullstack-devs 

It makes more sense to have this method be static as it'll be easier for the developer to use instead of having to rely on an instance. Also matches what iOS has, which is a static class method.


wangjoshuah:
lgtm

jophde:
This change is probably fine since it's a pure function.  I highly recommend avoiding making things static that aren't pure functions.  It will bite you.

mikeng13:
Sounds good thanks @jophde. I am very weary of static functions as well. It just made sense for this case :)

PR #59, opened 2016-12-20
Fix various bugs and crashes.

mikeng13:
@optimizely/fullstack-devs 

This fixes a number of crashes and bugs that our beta customer pointed out.

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
@@ -106,16 +106,7 @@ class BackgroundWatchersCache {
 
     @Nullable
     private JSONObject load() throws JSONException {
-        String backGroundWatchersFile = null;
-        try {
-            backGroundWatchersFile = cache.load(BACKGROUND_WATCHERS_FILE_NAME);
-        } catch (FileNotFoundException e) {
-            logger.info("Creating background watchers file");
-            backGroundWatchersFile = "{}";
-        } catch (IOException e) {
-            logger.error("Unable to load background watchers file", e);
-        }
-
+        String backGroundWatchersFile = cache.load(BACKGROUND_WATCHERS_FILE_NAME);
         if (backGroundWatchersFile == null) {
             backGroundWatchersFile = "{}";
         }
@@ -128,11 +119,6 @@ class BackgroundWatchersCache {
     }
 
     private boolean save(String backgroundWatchersJson) {
-        try {
-            return cache.save(BACKGROUND_WATCHERS_FILE_NAME, backgroundWatchersJson);
-        } catch (IOException e) {
-            logger.error("Unable to save background watchers file", e);
-            return false;
-        }
+        return cache.save(BACKGROUND_WATCHERS_FILE_NAME, backgroundWatchersJson);
     }
 }
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
@@ -117,7 +117,7 @@ class BackgroundWatchersCache {
         }
 
         if (backGroundWatchersFile == null) {
-            return null;
+            backGroundWatchersFile = "{}";
         }
 
         return new JSONObject(backGroundWatchersFile);
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
@@ -109,6 +109,7 @@ class BackgroundWatchersCache {
         String backGroundWatchersFile = cache.load(BACKGROUND_WATCHERS_FILE_NAME);
         if (backGroundWatchersFile == null) {
             backGroundWatchersFile = "{}";
+            logger.info("Creating background watchers file {}.", BACKGROUND_WATCHERS_FILE_NAME);
         }
 
         return new JSONObject(backGroundWatchersFile);
@@ -119,6 +120,13 @@ class BackgroundWatchersCache {
     }
 
     private boolean save(String backgroundWatchersJson) {
-        return cache.save(BACKGROUND_WATCHERS_FILE_NAME, backgroundWatchersJson);
+        logger.info("Saving background watchers file {}.", BACKGROUND_WATCHERS_FILE_NAME);
+        boolean saved = cache.save(BACKGROUND_WATCHERS_FILE_NAME, backgroundWatchersJson);
+        if (saved) {
+            logger.info("Saved background watchers file {}.", BACKGROUND_WATCHERS_FILE_NAME);
+        } else {
+            logger.warn("Unable to save background watchers file {}.", BACKGROUND_WATCHERS_FILE_NAME);
+        }
+        return saved;
     }
 }
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
@@ -64,6 +64,10 @@ class DataFileClient {
                     logger.info("Requesting data file from {}", url);
                     urlConnection = client.openConnection(url);
 
+                    if (urlConnection == null) {
+                        return null;
+                    }
+
                     client.setIfModifiedSince(urlConnection);
 
                     urlConnection.setConnectTimeout(5 * 1000);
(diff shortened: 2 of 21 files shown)


test-app/src/main/java/com/optimizely/ab/android/test_app/MainActivity.java
```diff
@@ -47,7 +47,7 @@ protected void onCreate(Bundle savedInstanceState) {
         optimizelyManager = myApplication.getOptimizelyManager();
 
         // Load Optimizely from a compiled in data file
-        final OptimizelyClient optimizely = optimizelyManager.initialize(this, R.raw.data_file);
+        final OptimizelyClient optimizely = optimizelyManager.initialize(getApplicationContext(), R.raw.data_file);
```
jophde: It's fine to pass this here.  We already get the activity context internally in the manager.  Your previous change checking the type of the binder should be enough to stop the crash.  The onServiceConnected(...) method may be called multiple times since we are still using application context but you are filtering out the calls that we are not interested in now.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -485,6 +485,10 @@ public void onActivityDestroyed(Activity activity) {
         @Override
         public void onServiceConnected(ComponentName className,
                                        IBinder service) {
+            if (!(service instanceof DataFileService.LocalBinder)) {
```
jophde: This should handle the case of onServiceConnected(ComponentName, IBinder) getting called multiple times because of us binding with application context.

jophde: This should actually be enough.  We probably don't need to change the manager to use activity and service context.  This method returning early will cause no issues as long as it gets called at least once with the correct type of binder.

test-app/src/main/java/com/optimizely/ab/android/test_app/MainActivity.java
mikeng13: Oh this was not supposed to make it in. I was just trying to repro the issue. Will revert.

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventDAO.java
```diff
@@ -120,17 +135,24 @@ boolean removeEvent(long eventId) {
...
         String[] selectionArgs = {String.valueOf(eventId)};
-        // Issue SQL statement.
-        int numRowsDeleted = dbHelper.getWritableDatabase().delete(EventTable.NAME, selection, selectionArgs);
-
-        if (numRowsDeleted > 0) {
-            logger.info("Removed event with id {} from db", eventId);
-            return true;
-        } else {
-            logger.error("Tried to remove an event id {} that does not exist", eventId);
+
```
vraja2: log "Removing {} from db" similar to what we do on record creation

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventDAO.java
```diff
@@ -120,17 +135,24 @@ boolean removeEvent(long eventId) {
...
+            if (numRowsDeleted > 0) {
+                logger.info("Removed event with id {} from db", eventId);
+                return true;
+            } else {
+                logger.error("Tried to remove an event id {} that does not exist", eventId);
+            }
+
+            return numRowsDeleted > 0;
+        } catch (Exception e) {
+            logger.error("Could not open db.", e);
```
vraja2: Is this the right error to log for an exception when deleting?

jophde: We are going to just catch all exceptions for db and network I/O since runtime exceptions can be thrown and we don't know which ones they are.

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventClient.java
```diff
@@ -64,23 +69,18 @@ public Boolean execute() {
                         logger.error("Unexpected response from event endpoint, status: " + status);
                         return Boolean.FALSE;
                     }
-                } catch (IOException e) {
+                } catch (Exception e) {
```
jophde: An exception should never make it here since they are being caught the try/catch inside of the Client#openConnection(URL).

android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
```diff
@@ -117,7 +117,7 @@ private JSONObject load() throws JSONException {
         }
 
         if (backGroundWatchersFile == null) {
-            return null;
+            backGroundWatchersFile = "{}";
```
jophde: Clients of this code weren't expecting null right?

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventClient.java
mikeng13: Isn't it possible for exceptions to be thrown in lines 54-62? We should be catching those as well.

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventDispatcher.java
```diff
@@ -71,26 +71,33 @@ void dispatch(@NonNull Intent intent) {
...
+                long interval = getInterval(intent);
+                serviceScheduler.schedule(intent, interval);
+                saveInterval(interval);
+                logger.info("Scheduled events to be dispatched");
+            } else {
+                // Quit trying to dispatch events because their aren't any in storage
+                serviceScheduler.unschedule(intent);
+                logger.info("Unscheduled event dispatch");
+            }
+        } catch (Exception e) {
```
jophde: ServiceScheduler#schedlue(Intent, int) doesn't do any I/O.  If this is about the NPE and the PendingIntent I think we should just check the PendingIntent for null since that's more predictable.

shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
```diff
@@ -81,10 +81,16 @@ public void schedule(Intent intent, long interval) {
      * @hide
      */
     public void unschedule(Intent intent) {
-        PendingIntent pendingIntent = pendingIntentFactory.getPendingIntent(intent);
-        alarmManager.cancel(pendingIntent);
-        pendingIntent.cancel();
-        logger.info("Unscheduled {}", intent.getComponent().toShortString());
+        if (intent != null) {
+            try {
+                PendingIntent pendingIntent = pendingIntentFactory.getPendingIntent(intent);
```
jophde: We should just check if pendingIntent is null or not and return if it is instead of catching NPE

android-sdk/src/main/java/com/optimizely/ab/android/sdk/BackgroundWatchersCache.java
mikeng13: Yeah they weren't. Your tests caught this issue :)

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventDispatcher.java
mikeng13: We also had that crash where `.schedule` failed because of that missing `INTERACT_ACROSS_USER` permission.

shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
mikeng13: I think `alarmManager.cancel` was also throwing an exception in one of those logs we saw.

jophde: Cool

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventDispatcher.java
jophde: Ok cool

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventClient.java
jophde: Yea they could throw IOException.  I would just keep it as that.  The try/catch in Client should catch crazy exceptions.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/DataFileClient.java
```diff
@@ -80,7 +84,7 @@ public String execute() {
                         logger.error("Unexpected response from data file cdn, status: {}", status);
                         return null;
                     }
-                } catch (IOException e) {
+                } catch (Exception e) {
```
jophde: Ditto here, just keep it as IOException.

mikeng13:
Re-opening PR to do squash merge instead of just merge

PR #81, opened 2017-01-20
Release version 0.5.0

vraja2:
This is already on Bintray, but want to make an official intermediate release prior to GA.

@optimizely/fullstack-devs


vraja2:
also, had to merge 'master' into 'devel' since some previous releases squashed and merged rather than just merged into master.

PR #89, opened 2017-01-24
Fix alignment of variations in test-app

elliotykim:
The text of the variations didn't line up and looked weird in screenshots


elliotykim:
@vraja2 @jprovine @optimizely/fullstack-devs

PR #90, opened 2017-01-24
Elliot/test app alignment

elliotykim:
@vraja2 if at first you don't succeed


elliotykim:
Already approved here:
https://github.com/optimizely/android-sdk/pull/89

PR #119, opened 2017-07-13
OASIS-1501

elliotykim:
Move close() to finally

@optimizely/fullstack-devs @mikeng13 @thomaszurkan-optimizely


elliotykim:
Tom already made this fix in another PR

PR #126, opened 2017-07-17
Handle errors so apps don't crash

elliotykim:
OASIS-1516

@optimizely/fullstack-devs


javierarboleda:
If you're now catching any error or exception, onStart is not called when there's an error or exception thrown. This would be an issue if anyone is doing any significant logic or navigation calls, such as what you provide in your developers example code where you have an activity starting from the onStart method:


    optimizelyManager.initialize(this, new OptimizelyStartListener()  {
      @Override public void onStart(OptimizelyClient optimizely) {
          assert optimizely == optimizelyManager.getOptimizely();
          Intent intent = new Intent(FirstActivity.this, SubsequentActivity.class);
          startActivity(intent);
      }
    });

A solution here can be to add an onFailure call to the OptimizelyStartListener.

thomaszurkan-optimizely:
Good idea.  Can you capture this in the issue that you opened https://github.com/optimizely/android-sdk/issues/147?  

I've been trying to reproduce and I've been unable to do so.  Unfortunately, we only have API 17 on emulator.   Are you experiencing the exception in the emulator as well?

PR #160, opened 2017-10-27
update for android o after device tests.  updated the change log and …

thomaszurkan-optimizely:
…a litle of the javadoc for initialize


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -120,7 +120,7 @@ private void notifyStartListener() {
     }
 
     /**
-     * Initialize Optimizely Synchronously
+     * Initialize Optimizely Synchronously and downloads the latest datafile in the background to cache.
```
elliotykim: Grammar parallelism nit: since "Initialize" is in the imperative form, "download" should also be imperative to match

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -150,7 +150,8 @@ public OptimizelyClient initialize(@NonNull Context context, @NonNull String dat
     }
 
     /**
-     * Initialize Optimizely Synchronously
+     * Initialize Optimizely Synchronously. Load the resource, use it to initialize Optimizely, do a download of the datafile
+     * in the background to cache.
```
elliotykim: This description should be worded similarly to the one above for `initialize(@NonNull Context context, @NonNull String datafile)` except for a part about what datafile is being used to initialize synchronously at first.

At the moment, "and downloads the latest datafile in the background to cache" and "Load the resource, use it to initialize Optimizely, do a download of the datafile in the background to cache." are vastly differently.

We should make it clear that if you use this method, it will ignore the cached datafile and initialize using the datafile you pass into this call. To check if a cached datafile exists, they should check with `isDatafileCached()`.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -173,7 +174,7 @@ public OptimizelyClient initialize(@NonNull Context context, @RawRes int datafil
     }
 
     /**
-     * Initialize Optimizely Synchronously
+     * Initialize Optimizely Synchronously.  This one uses the cached datafile and expects it to exist.
```
elliotykim: We could mention `isDatafileCached()` here as well

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
```diff
@@ -49,26 +53,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
-//                }
-//            }
-//
-//            startActivity(intent);
-//
-//            return;
-//        }
 
+        // Example of initialize from raw file and use cached file after that.
+//        DatafileHandler datafileHandler = optimizelyManager.getDatafileHandler();
```
elliotykim: `OptimizelyManager#isDatafileCached()` is the shortcut for this

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
thomaszurkan-optimizely: added mention of isDatafileCached.

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
```diff
@@ -49,26 +53,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
-//            return;
-//        }
 
+        // Example of initialize from raw file and use cached file after that.
+        if (false) {
+
+            if (optimizelyManager.isDatafileCached(myApplication)) {
+                optimizelyManager.initialize(myApplication);
+            } else {
+                optimizelyManager.initialize(myApplication, R.raw.data_file);
```
elliotykim: you could also rename the resource file to `datafile.json` to be consistent with the naming convention used in the code

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
```diff
@@ -49,26 +53,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
-//                }
-//            }
-//
-//            startActivity(intent);
-//
-//            return;
-//        }
 
+        // Example of initialize from raw file and use cached file after that.
+        if (false) {
```
elliotykim: at first glance, `if (false) {...}` looked like a typo, but i see what you're doing now.

@wangjoshuah had a suggestion for handling this which i'll amend slightly. could you declare a variable that gates between this block and the async initialization code below?

it could either be a constant boolean named something like `INITIALIZE_ASYNCHRONOUSLY` that's checked by if-else or an enum that's checked by a switch statement for the 2 cases.

this allows a developer running the test app to run either code block by changing the value of the variable and makes the code look more intentional.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -120,13 +120,15 @@ private void notifyStartListener() {
     }
 
     /**
-     * Initialize Optimizely Synchronously
+     * Initialize Optimizely Synchronously using the datafile pass in while downloading the latest datafile in the background from the CDN to cache.
```
elliotykim: passed in

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -120,13 +120,15 @@ private void notifyStartListener() {
...
+     * It should be noted that even though it initiates a download of the datafile to cache. This method does not use that cached datafile.
```
elliotykim: these 2 sentences should be combined into a single sentence:
"It should be noted that even though it initiates a download of the datafile to cache, this method does not use that cached datafile."

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -150,7 +152,10 @@ public OptimizelyClient initialize(@NonNull Context context, @NonNull String dat
     }
 
     /**
-     * Initialize Optimizely Synchronously
+     * Initialize Optimizely Synchronously by loading the resource, use it to initialize Optimizely,
+     * and downloading the latest datafile from the CDN in the background to cache.
+     * It should be noted that even though it initiates a download of the datafile to cache. This method does not use that cached datafile.
```
elliotykim: combine sentences

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
elliotykim: single space between sentences

PR #167, opened 2017-11-23
Feature - Initialization Sync, Async 

mnoman09:


msohailhussain:
Sure, we'll update the PR.

mnoman09:
@thomaszurkan-optimizely I have updated the PR as per requirement. you can review the code. If you have any further query, notify us.

android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyManagerTest.java
```diff
@@ -132,6 +115,53 @@ public void initializeInt() {
...
+    @Test
+    public void initializeSyncWithEmptyDatafile() {
+        //for this case to pass empty the data file or enter any garbage data given on R.raw.emptydatafile this path
+        Context context = mock(Context.class);
+        Context appContext = mock(Context.class);
+        when(context.getApplicationContext()).thenReturn(appContext);
+        when(appContext.getPackageName()).thenReturn("com.optly");
+        optimizelyManager.initialize(InstrumentationRegistry.getTargetContext(), R.raw.emptydatafile);
+        verify(logger).error(eq("Unable to parse compiled data file"), any(ConfigParseException.class));
+    }
```
thomaszurkan-optimizely: Can we add an assert that optimizelyManager.getOptimizely.isValid is false?

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -207,45 +193,32 @@ public OptimizelyClient initialize(@NonNull Context context) {
...
-                if (datafile != null) {
+                if (datafile != null && !datafile.isEmpty()) {
                     injectOptimizely(context, userProfileService, datafile);
                 } else {
-                    // We should always call the callback even with the dummy
-                    // instances.  Devs might gate the rest of their app
-                    // based on the loading of Optimizely
+                    //if datafile is null than it should be able to take from cache and if not present
+                    //in Cache than should be able to get from raw data file
+                    optimizelyClient= initialize(context,datafileRes);
```
thomaszurkan-optimizely: So, the only problem here is that initialize(context, datafileRes) will eventually call initialize(context, datafile) which does a loadDatafile so there is an extra datafile download every async initialize .

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -207,45 +193,32 @@ public OptimizelyClient initialize(@NonNull Context context) {
...
 
-    DatafileLoadedListener getDatafileLoadedListener(final Context context) {
+    DatafileLoadedListener getDatafileLoadedListener(final Context context, @RawRes final int datafileRes) {
         return new DatafileLoadedListener() {
             @RequiresApi(api = Build.VERSION_CODES.HONEYCOMB)
             @Override
             public void onDatafileLoaded(@Nullable String datafile) {
                 // App is being used, i.e. in the foreground
-                if (datafile != null) {
+                if (datafile != null && !datafile.isEmpty()) {
```
thomaszurkan-optimizely: The datafile will be empty "" if you get a 304.  So, you would be doing 2 https requests for one datafile if the datafile hasn't changed.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -207,45 +193,32 @@ public OptimizelyClient initialize(@NonNull Context context) {
...
-    public void initialize(@NonNull Context context, @NonNull OptimizelyStartListener optimizelyStartListener) {
+    @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+    public void initialize(@NonNull final Context context, @RawRes final int datafileRes, @NonNull OptimizelyStartListener optimizelyStartListener) {
         if (!isAndroidVersionSupported()) {
             return;
         }
-        this.optimizelyStartListener = optimizelyStartListener;
-        datafileHandler.downloadDatafile(context, projectId, getDatafileLoadedListener(context));
+        setOptimizelyStartListener(optimizelyStartListener);
+        datafileHandler.downloadDatafile(context, projectId,getDatafileLoadedListener(context,datafileRes));
```
thomaszurkan-optimizely: This call, calls DatafileLoader.loadDatafile which tries to load from both cache and cdn.  It returns whichever returns first.   I think that logic may have to change slightly as well.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
mnoman09: it's solved now.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
mnoman09: Done i made changes in datafileloader.java and fixed it

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileLoader.java
```diff
@@ -68,13 +68,13 @@ public void getDatafile(@NonNull String datafileUrl, @Nullable DatafileLoadedLis
                         this,
                         datafileLoadedListener,
                         logger);
-        LoadDatafileFromCacheTask loadDatafileFromCacheTask =
-                new LoadDatafileFromCacheTask(datafileCache,
-                        this,
-                        datafileLoadedListener);
+//        LoadDatafileFromCacheTask loadDatafileFromCacheTask =
```
thomaszurkan-optimizely: Why is this commented out?  Should it be removed?

thomaszurkan-optimizely: Can we go ahead and refactor this method so that it takes a parameter with default == to false (fromCache)?  If from cache we can try both or just try from cache. Thanks!

mnoman09: I think refactoring this logic will have bigger impact than just removing it because datafileloadedlistener is notified only once which is either from cache or from downloaded file which ever returns first. we are already handling getFromCache logic which works for both synchronous and asynchronous in optimizelymanager so I don't think there is any need of handling this here. What do you think?

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileLoader.java
```diff
@@ -163,10 +129,7 @@ protected void onPostExecute(@Nullable String dataFile) {
             // Only send null or a real datafile
             // If the datafile is empty it means we got a 304
             // We should have already sent the local datafile in this case
```
thomaszurkan-optimizely: almost there!  can you update the comment as well?

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -146,18 +146,17 @@ public OptimizelyClient initialize(@NonNull Context context, @NonNull String dat
             logger.error("Unable to build OptimizelyClient instance", e);
         }
 
-        datafileHandler.downloadDatafile(context, projectId, null);
```
thomaszurkan-optimizely: rather than remove this download and break any existing apps that might be expecting a download to occur here, can i suggest we create two mehtods:

initialize(Context context, String datafile, downloadToCache)
and have 
initialize(Context context, String datafile) {
  initialize(Context, datafile, true);
}
The first initialize could even be protected.  we would use that passing in false inside the listener if we got back null or empty string

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -167,38 +166,38 @@ public OptimizelyClient initialize(@NonNull Context context, @NonNull String dat
...
-        String datafile = datafileHandler.loadSavedDatafile(context, projectId);
-
-        if (datafile != null) {
-            return initialize(context, datafile);
+    public String getDatafile(Context context,@RawRes int datafileRes){
+     try {
+        if (isDatafileCached(context)) {
+            return datafileHandler.loadSavedDatafile(context, projectId);
+        } else {
+            return loadRawResource(context, datafileRes);
```
thomaszurkan-optimizely: is there a way we could make this an Integer so that we can check if it is null rather than rely on an exception if the user decides they don't want a datafile in the resource?

msohailhussain:
@thomaszurkan-optimizely Can you please review. Latest checked-in with all changes https://github.com/optimizely/android-sdk/pull/167/commits/7781af34aa068e29c37e694fd761f6ee4631e3a0

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -207,53 +231,38 @@ public OptimizelyClient initialize(@NonNull Context context) {
...
-    /**
      * This method does the same thing except it can be used with a generic {@link Context}.
      * @param context                 any type of context instance
      * @param optimizelyStartListener callback that {@link OptimizelyClient} instances are sent to.
-     * @see #initialize(Activity, OptimizelyStartListener)
+     * @see #initialize(Context, int ID, OptimizelyStartListener)
      */
-    public void initialize(@NonNull Context context, @NonNull OptimizelyStartListener optimizelyStartListener) {
+    @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+    public void initialize(@NonNull final Context context, @RawRes final int datafileRes, @NonNull OptimizelyStartListener optimizelyStartListener) {
```
thomaszurkan-optimizely: Sorry, I missed this.  Can we also allow null for the resource here?

PR #168, opened 2017-12-08
- Updated Java sdk version from 1.8 to 2.0.0-beta2 and removed Live variables

mnoman09:
* - Updated Java sdk version from 1.8 to 2.0.0-beta2
- Removed
   getVariableString
   getVariableBoolean
   getVariableInteger
   getVariableDouble
-Remove track with revenue as a parameter. Pass the revenue value as an event tag instead
   track(String, String, long)
   track(String, String, Map<String, String>, long)

* added multidex enabled support

* make common Constant of $multidex_ver to use in all gradle


msohailhussain:
@thomaszurkan-optimizely we are investigating why test_api_version_25 getting failed. I will let you know once its fixed. It's showing out of memory but need to fix it.

android-sdk/build.gradle
```diff
@@ -62,7 +62,9 @@ dependencies {
     compile project(':datafile-handler')
     compile project(':event-handler')
     compile project(':user-profile')
-    compile "com.optimizely.ab:core-api:$java_core_ver"
+    compile ("com.optimizely.ab:core-api:$java_core_ver") {
+        exclude group: 'com.google.code.findbugs'
```
wangjoshuah: why are we excluding the find bugs scan?

test-app/src/main/java/com/optimizely/ab/android/test_app/MyApplication.java
```diff
@@ -68,8 +67,8 @@ public String getAnonUserId() {
     @Override
     public void onCreate() {
         super.onCreate();
-
-        // This app is built against a real Optimizely project with real experiments set.  Automated
+        
+      // This app is built against a real Optimizely project with real experiments set.  Automated
```
wangjoshuah: *nit*
undo this part

android-sdk/build.gradle
mnoman09: As discussed in PR #170 
https://github.com/optimizely/android-sdk/pull/170/files/68ee83ccbd479369214c5ea08e21ed6b6acbec2c

wangjoshuah: thanks. we should move off of `findbugs` anyways since they are not currently being supported 
@mnoman09 what other static code linters should we use?

PR #184, opened 2018-03-12
get enabled feature pr

mnoman09:
* Added Wrapper for getEnabledFeatures 
* Added its unit tests
* updated java version to 2.0.0-beta6
* updated datafile and added featureEnabled as it was not passing all unittests of due to featureEnabled was not set

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,7 @@ ext {
     build_tools_version = "27.0.0"
     min_sdk_version = 10
     target_sdk_version = 26
-    java_core_ver = "2.0.0-beta4"
+    java_core_ver = "2.0.0-beta6"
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -53,7 +53,7 @@ ext {
     build_tools_version = "27.0.0"
     min_sdk_version = 10
     target_sdk_version = 26
-    java_core_ver = "2.0.0-beta4"
+    java_core_ver = "2.0.0-beta6"
```
mikeng13: Any reason we are jumping from 4 to 6?

mnoman09: yes, we changed it to beta6 because getEnabledFeatures feature was included in java sdk beta6 release so in order to write its wrapper we had to update it.

PR #195, opened 2018-04-19
fix for blue apron by providing default logger if none is supplied.

thomaszurkan-optimizely:


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -628,7 +629,323 @@ public OptimizelyManager build(Context context) {
                 try {
                     logger = LoggerFactory.getLogger(OptimizelyManager.class);
                 } catch (Exception e) {
-                    logger = LoggerFactory.getLogger("com.optimizely.ab.android.sdk.OptimizelyManager");
+                    try {
+                        logger = LoggerFactory.getLogger("com.optimizely.ab.android.sdk.OptimizelyManager");
+                    }
+                    catch (Exception e1) {
+                        logger = new Logger() {
```
mikeng13: This is crazy haha. Can we create a class that encapsulates this and pull it in here?

PR #196, opened 2018-04-24
update to load the user profile service asynch and synch. remove inva…

thomaszurkan-optimizely:
…lid experiments if started asynch (latest datafile) or used cached datafile.

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
@@ -215,7 +215,7 @@ public class OptimizelyManager {
     }
 
     private void cleanupUserProfileCache(UserProfileService userProfileService) {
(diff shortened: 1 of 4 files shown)


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -200,6 +214,31 @@ public OptimizelyClient initialize(@NonNull Context context, @RawRes Integer dat
...
+        else {
+            return;
+        }
+
+        ProjectConfig config = optimizelyClient.getProjectConfig();
+        if (config == null) {
+            return;
+        }
+
+        final Set<String> experimentIds = config.getExperimentIdMapping().keySet();
```
mikeng13: Not sure if this operation is cause for concern, but given enough experiments it could hang the main thread for a bit. Can we do this in the worker thread instead?

PR #199, opened 2018-05-25
fix for datafile download start when datafile has not changed or when…

thomaszurkan-optimizely:
… starting synchronously


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -303,8 +309,7 @@ public void onDatafileLoaded(@Nullable String datafile) {
                 } else {
                     //if datafile is null than it should be able to take from cache and if not present
                     //in Cache than should be able to get from raw data file
-                    optimizelyClient = initialize(context,getDatafile(context,datafileRes),false);
-                    notifyStartListener();
+                    injectOptimizely(context, userProfileService, getDatafile(context,datafileRes));
```
mikeng13: Ah nice catch here.

PR #200, opened 2018-06-06
bring back getVariable(String,Integer,Double,Boolean)

thomaszurkan-optimizely:
unit tests added as well.

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
@@ -624,13 +624,13 @@ public class OptimizelyClient {
         }
 
         if (variableKey == null || variableKey.isEmpty()) {
-            logger.warn("Invalid live variable key (null or empty) " +
+            logger.error("Invalid live variable key (null or empty) " +
                     "for user {}", userId);
             return null;
         }
 
         if (userId == null || userId.isEmpty()) {
-            logger.warn("Invalid userId (null or empty) " +
+            logger.error("Invalid userId (null or empty) " +
                     "for live variable {}", variableKey);
             return null;
         }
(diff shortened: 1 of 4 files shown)


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
```diff
@@ -551,4 +557,226 @@ public NotificationCenter getNotificationCenter() {
...
+     * If {@link RaiseExceptionErrorHandler} is provided, either a live variable is returned, or an exception is
+     * thrown.
+     * If {@link NoOpErrorHandler} is used, either a live variable or {@code null} is returned.
+     *
+     * @param projectConfig the current project config
+     * @param variableKey the key for the live variable being retrieved from the current project config
+     * @return the live variable to retrieve for the given variable key
+     *
+     */
+    private LiveVariable getLiveVariable(ProjectConfig projectConfig, String variableKey) {
```
mikeng13: I think we should null-check the `variableKey` since it is a user-provided input or we can do it at a higher level, like the caller of this function

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
```diff
@@ -551,4 +557,226 @@ public NotificationCenter getNotificationCenter() {
...
+     * @return String value of the live variable
+     */
+    @Deprecated
+    public @Nullable
+    String getVariableString(@NonNull String variableKey,
+                             @NonNull String userId,
+                             @NonNull Map<String, String> attributes,
+                             boolean activateExperiment) {
+
+        if (!isValid()) {
```
mikeng13: Let's null-check the `userId` and `variableKey` input params here

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
```diff
@@ -551,4 +557,237 @@ public NotificationCenter getNotificationCenter() {
...
+            logger.warn("Optimizely is not initialized, could not get live variable {} " +
+                    "for user {}", variableKey, userId);
+            return null;
+        }
+
+        if (variableKey == null || variableKey.isEmpty()) {
+            logger.warn("Invalid live variable key (null or empty) " +
```
mikeng13: I'd make the argument for making these `error` logs so the developer will be more inclined to fix it or null-check on their end before passing to us.

PR #205, opened 2018-06-20
Upgrade to Java 2.1.1 which fixes feature impression sending for vari…

mikeng13:
…ations toggled off.


mikeng13:
@aliabbasrizvi  I manually tested that the variation with the feature toggled OFF will send the event over to our event endpoint and that the impression shows up correctly in the results page.

PR #207, opened 2018-07-11
add new sdkKey cdn location and allow the datafileconfig to be passed…

thomaszurkan-optimizely:
… in overriding host if need be

### DIFF
--- a/shared/src/main/java/com/optimizely/ab/android/shared/DatafileConfig.java
+++ b/shared/src/main/java/com/optimizely/ab/android/shared/DatafileConfig.java
@@ -31,19 +31,19 @@ public class DatafileConfig {
     public static String defaultHost = "https://cdn.optimizely.com";
     public static String projectUrlSuffix = "/json/%s.json";
     public static String environmentUrlSuffix = "/datafiles/%s.json";
-    public static String delimitor = "::::";
+    public static String delimiter = "::::";
 
     private final String projectId;
     private final String sdkKey;
     private final String host;
-    private final String environmentUrl;
-    private final String projectIdUrl;
+    private final String datafileUrlString;
 
     /**
      * Constructor used to construct a DatafileConfig to get cache key, url,
      * for the appropriate environment.  One or the other can be null.  But, not both.
      * @param projectId project id string.
      * @param sdkKey the environment url.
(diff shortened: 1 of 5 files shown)


shared/src/main/java/com/optimizely/ab/android/shared/DatafileConfig.java
```diff
@@ -28,22 +28,52 @@
...
 
     /**
      * Constructor used to construct a DatafileConfig to get cache key, url,
      * for the appropriate environment.  One or the other can be null.  But, not both.
      * @param projectId project id string.
      * @param sdkKey the environment url.
      */
-    public DatafileConfig(String projectId, String sdkKey) {
+    public DatafileConfig(String projectId, String sdkKey, String host) {
         assert(projectId != null || sdkKey != null);
```
mikeng13: Can this crash and throw? We are making this class public so we should document that it throws and also try/catch around where we use it in the manager

shared/src/main/java/com/optimizely/ab/android/shared/DatafileConfig.java
```diff
@@ -28,22 +28,52 @@
...
         this.projectId = projectId;
         this.sdkKey = sdkKey;
+        this.host = host;
+
+        if (sdkKey != null) {
+            this.environmentUrl = String.format((this.host + environmentUrlSuffix), sdkKey);
```
mikeng13: Can we use just 1 variable, `datafileUrl`, which gets set here and then in `getUrl` we can just return that url instead of checking for the `sdkKey` there as well

shared/src/main/java/com/optimizely/ab/android/shared/DatafileConfig.java
thomaszurkan-optimizely: we do not use this method in the manager.  it is only used to override the datafile config.  it is not documented because it is essentially a testing mechanism.

shared/src/main/java/com/optimizely/ab/android/shared/DatafileConfig.java
thomaszurkan-optimizely: makes sense.

PR #222, opened 2018-09-07
Add @Nullable annotation to getNotificationCenter()

fingertricks:
In circumstances where initialisation fails, ``OptimizelyManager`` returns a dummy ``OptimizelyClient`` which has a null ``NotificationCenter``.

Adding this annotation warns SDK users that they shouldn’t rely on a ``@NonNull NotificationCenter`` being returned, for example if they call ``addActivateNotificationListener()`` after ``initialise()`` This bit a client of mine in production so hoping this will help other users.

Android Studio now displays appropriate warnings when calling this API from Java, including from the test app. In Kotlin though the following would not compile if this annotation is added:

``optimizelyClient.notificationCenter.addActivateNotificationListener(...)``
``Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type NotificationCenter``

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
@@ -546,9 +546,11 @@ public class OptimizelyClient {
     /**
      * Return the notification center {@link NotificationCenter} used to add notifications for events
      * such as Activate and track.
-     * @return
+     * @return The {@link NotificationCenter} or null if Optimizely is not initialized (or
+     * initialization failed)
      */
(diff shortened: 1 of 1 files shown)


thomaszurkan-optimizely:
@fingertricks can you sign a CLA?  

https://github.com/optimizely/java-sdk/blob/master/CONTRIBUTING.md
https://docs.google.com/a/optimizely.com/forms/d/e/1FAIpQLSf9cbouWptIpMgukAKZZOIAhafvjFCV8hS00XJLWQnWDFtwtA/viewform

After that, we will merge it in.

fingertricks:
CLA signed 👍

Issue #234, opened 2018-10-29
Crash with IllegalStateException - Apps may not schedule more than 100 distinct jobs


benvonhandorf:
We're seeing a crash in the Optimizely stack around scheduling jobs with the JobScheduler.  The app also makes use of JobScheduler, via the WorkManager library, but the app currently only schedules 3-5 jobs, maximum.

These crashes are mostly happening in the background, according to fabric.

Relevant library versions:

implementation('com.optimizely.ab:android-sdk:2.1.0')
android.arch.work:work-runtime: 1.0.0-alpha10
android.arch.work:work-firebase: 1.0.0-alpha10
com.firebase:firebase-jobdispatcher:0.8.5

```
Fatal Exception: java.lang.IllegalStateException: Apps may not schedule more than 100 distinct jobs
       at android.os.Parcel.readException(Parcel.java:1967)
       at android.os.Parcel.readException(Parcel.java:1905)
       at android.app.job.IJobScheduler$Stub$Proxy.schedule(IJobScheduler.java:180)
       at android.app.JobSchedulerImpl.schedule(JobSchedulerImpl.java:44)
       at com.optimizely.ab.android.shared.ServiceScheduler.setRepeating(ServiceScheduler.java:134)
       at com.optimizely.ab.android.shared.ServiceScheduler.schedule(ServiceScheduler.java:83)
       at com.optimizely.ab.android.datafile_handler.DefaultDatafileHandler.startBackgroundUpdates(DefaultDatafileHandler.java:122)
       at com.optimizely.ab.android.sdk.OptimizelyManager.injectOptimizely(OptimizelyManager.java:442)
       at com.optimizely.ab.android.sdk.OptimizelyManager$2.onDatafileLoaded(OptimizelyManager.java:343)
       at com.optimizely.ab.android.datafile_handler.DefaultDatafileHandler$1.onDatafileLoaded(DefaultDatafileHandler.java:78)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader.notify(DatafileLoader.java:81)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader.access$000(DatafileLoader.java:35)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader$RequestDatafileFromClientTask.onPostExecute(DatafileLoader.java:134)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader$RequestDatafileFromClientTask.onPostExecute(DatafileLoader.java:85)
[...]

PR #236, opened 2018-11-08
(fix)wrap schedule in a try catch in case of too many scheduled jobs which…

thomaszurkan-optimizely:
… throws an exception

wrap the job scheduler schedule in a try catch.  Customer is experiencing difficulty starting in the background and trying to schedule another background update causes a Illegal State Exception.  

This is not necessarily the SDK's fault but the exception should not stop the app.  So, we are wrapping the call with a try catch.

https://github.com/optimizely/android-sdk/issues/234

### DIFF
--- a/shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
+++ b/shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
@@ -131,10 +131,14 @@ public class ServiceScheduler {
 
             builder.setExtras(persistableBundle);
 
-            if (jobScheduler.schedule(builder.build()) != RESULT_SUCCESS) {
-                logger.error("ServiceScheduler", "Some error while scheduling the job");
+            try {
+                if (jobScheduler.schedule(builder.build()) != RESULT_SUCCESS) {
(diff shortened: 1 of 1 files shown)


shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
```diff
@@ -131,10 +131,14 @@ private void setRepeating(long interval, PendingIntent pendingIntent, Intent int
...
 
-            if (jobScheduler.schedule(builder.build()) != RESULT_SUCCESS) {
-                logger.error("ServiceScheduler", "Some error while scheduling the job");
+            try {
+                if (jobScheduler.schedule(builder.build()) != RESULT_SUCCESS) {
+                    logger.error("ServiceScheduler", "Some error while scheduling the job");
+                }
+            }
+            catch (Exception e) {
+                logger.error("Problem scheduling job ", e);
```
mikeng13: We should probably provide more context on this error, like the fact that the datafile sync scheduling failed.

thomaszurkan-optimizely: I added the component name which would in this case be DatafileService

Issue #258, opened 2019-03-01
Proguard fails on latest version of sdk 


matt-harris-mc:
Running the latest version (3.0) of the sdk fails when proguard is enabled with the below: 

`
Warning: com.optimizely.ab.config.parser.JacksonConfigParser$ProjectConfigModule: can't find referenced method 'com.fasterxml.jackson.databind.module.SimpleModule addDeserializer(java.lang.Class,com.fasterxml.jackson.databind.JsonDeserializer)' in program class com.optimizely.ab.config.parser.JacksonConfigParser$ProjectConfigModule
`

aliabbasrizvi:
Thanks for reporting @matt-harris-mc . We are looking into this.

thomaszurkan-optimizely:
hi @matt-harris-mc what proguard rules are you using?

matt-harris-mc:
We have a set of rules locally that we use for other stuff, but I was under the impression any required rules are automatically included in the aar. Are you worried that we possibly have rules that conflict with those?

thomaszurkan-optimizely:
can you try the following:
can you add the following to your proguard rules?
```
-keep class com.fasterxml.jackson.** {*;}
```

matt-harris-mc:
@thomaszurkan-optimizely I added that line to my proguard file and it is still failing with the same message. Also reverted to 2.1.4 to confirm my build is still working and it is not something else local to me.

thomaszurkan-optimizely:
so, 2.1.4 works and 3.0 doesn't?

matt-harris-mc:
Yep 2.1.4 is building without issue.

thomaszurkan-optimizely:
compile group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: '2.9.8'
are you compiling in jackson-databind?

matt-harris-mc:
We are not using it directly or even indirectly. It looks like the only reference to jackson we have is in a progaurd section we have included (commented as for firebase): `-keepnames class com.fasterxml.jackson.** { *; }`

thomaszurkan-optimizely:
hmm, if you add the jackson-databind in there i would assume it would work.  we are looking into a solution.  thanks!

matt-harris-mc:
Yea it looks like adding that lib as a dependency does fix the issue. Going to hold on 2.1.4 for now though. Thanks for the quick responses.

thomaszurkan-optimizely:
[...]

PR #266, opened 2019-04-19
(fix): this fixes proguard rules when running minifyEnabled

thomaszurkan-optimizely:
## Summary
I was able to reproduce this problem turning on minify on a kotlin project.  The following rules and compile cleanups, enabled the apk to be created, installed, and ran, without issue. 

## Test plan
Created test apps to reproduce the issue.
 
## Issues
- #258

### DIFF
--- a/android-sdk/build.gradle
+++ b/android-sdk/build.gradle
@@ -66,7 +66,6 @@ dependencies {
         exclude group: 'com.google.code.findbugs'
     }
 
-    compileOnly group: 'org.slf4j', name: 'slf4j-android', version: '1.7.25'
     compileOnly "com.fasterxml.jackson.core:jackson-databind:$jacksonversion"
     implementation "com.android.support:support-annotations:$support_annotations_ver"
 
--- a/android-sdk/build.gradle
+++ b/android-sdk/build.gradle
@@ -67,7 +67,7 @@ dependencies {
     }
 
     compileOnly group: 'org.slf4j', name: 'slf4j-android', version: '1.7.25'
-
+    compileOnly "com.fasterxml.jackson.core:jackson-databind:$jacksonversion"
     implementation "com.android.support:support-annotations:$support_annotations_ver"
 
     testImplementation "junit:junit:$junit_ver"
--- a/shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
+++ b/shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
@@ -157,11 +157,11 @@ public class ServiceScheduler {
             Integer id = null;
             try {
                 id = (Integer) Class.forName(clazz).getDeclaredField("JOB_ID").get(null);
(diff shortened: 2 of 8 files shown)


thomaszurkan-optimizely:
@mnoman9 can you share your project with me? How did you install the aar s?

thomaszurkan-optimizely:
@mnoman09 after looking at your project and changing it to use my branch instead of master for the fix, it worked.

mnoman09:
yes I tested it with my project and It's working so lgtm.

shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
```diff
@@ -157,11 +157,11 @@ private void cancelRepeating(PendingIntent pendingIntent, Intent intent) {
             Integer id = null;
             try {
                 id = (Integer) Class.forName(clazz).getDeclaredField("JOB_ID").get(null);
-                pendingIntent.cancel();
                 // only cancel periodic services
                 if (ServiceScheduler.isScheduled(context, id)) {
                     jobScheduler.cancel(id);
                 }
+                pendingIntent.cancel();
```
mikeng13: What does moving this here do?

android-sdk/build.gradle
```diff
@@ -66,8 +66,7 @@ dependencies {
         exclude group: 'com.google.code.findbugs'
     }
 
-    compileOnly group: 'org.slf4j', name: 'slf4j-android', version: '1.7.25'
```
mikeng13: We don't need this anymore?

android-sdk/build.gradle
```diff
@@ -66,8 +66,7 @@ dependencies {
...
-
+    compileOnly "com.fasterxml.jackson.core:jackson-databind:$jacksonversion"
```
mikeng13: Does this mean we explicitly bundle jackson with our SDK? Would the user be able to exclude it from their app?

shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
thomaszurkan-optimizely: It is just consistent with what is below it.  We first cancel the alarm and then the pendingIntent.  Here we cancel the job and then the pending intent.  Basically, it is so if the pendingIntent throws, we have still cancelled the scheduled job.

android-sdk/build.gradle
thomaszurkan-optimizely: We have a compile only for noveo android logger.  We don't need both.

android-sdk/build.gradle
thomaszurkan-optimizely: It is compile only.  So, it is not a explicit dependency.  If you are not using jackson, then, there is no effect.  If you are using jackson, you would add this as a implementation in your gradle.  Essentially, this is so our jackson code compiles.  We still default to gson.

Issue #267, opened 2019-04-22
IntentService -- OutOfMemoryError


jia-tse-zocdoc:
Hi, 

I'm seeing this error, happening only on Android 8 (Oreo) devices. 

```
Fatal Exception: java.lang.OutOfMemoryError: pthread_create (1040KB stack) failed: Try again
       at java.lang.Thread.nativeCreate(Thread.java)
       at java.lang.Thread.start(Thread.java:733)
       at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:970)
       at java.util.concurrent.ThreadPoolExecutor.processWorkerExit(ThreadPoolExecutor.java:1038)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1180)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
       at java.lang.Thread.run(Thread.java:764)
```

The crash console is showing there are 1177 threads created (doubt this is intentional!). And most of the threads look like this: 

```
IntentService[EventHandlerService]
       at android.os.MessageQueue.nativePollOnce(MessageQueue.java)
       at android.os.MessageQueue.next(MessageQueue.java:325)
       at android.os.Looper.loop(Looper.java:142)
       at android.os.HandlerThread.run(HandlerThread.java:65)
```

I'm not entirely sure this is an error from the optimizely SDK, but I saw you have this file, so it might be related. 

https://github.com/optimizely/android-sdk/blob/992acca918557031e4b1289277cc54976d5f945b/event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventIntentService.java#L53

matt-harris-mc:
We are also seeing similar crashes with a slightly different stack trace. 
```
Fatal Exception: java.lang.OutOfMemoryError: pthread_create (1040KB stack) failed: Out of memory
       at java.lang.Thread.nativeCreate(Thread.java)
       at java.lang.Thread.start(Thread.java:733)
       at java.util.concurrent.ThreadPoolExecutor.addWorker(ThreadPoolExecutor.java:970)
       at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1377)
       at android.os.AsyncTask.executeOnExecutor(AsyncTask.java:651)
       at com.optimizely.ab.android.shared.ScheduledJobService.onStartJob(ScheduledJobService.java:145)
[...]

PR #268, opened 2019-04-22
(fix):potential fix for multi scheduling of event work

thomaszurkan-optimizely:
## Summary
- Problems reported with EventScheduler scheduling too many jobs.

## Issues
- #267

### DIFF
--- a/event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventRescheduler.java
+++ b/event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventRescheduler.java
@@ -101,8 +101,10 @@ public class EventRescheduler extends BroadcastReceiver {
                     // with wifi the service will be rescheduled on the interval.
                     // Wifi connection state changes all the time and starting services is expensive
                     // so it's important to only do this if we have stored events.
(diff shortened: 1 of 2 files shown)


event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventRescheduler.java
```diff
@@ -101,8 +101,10 @@ void reschedule(@NonNull Context context, @NonNull Intent broadcastIntent, @NonN
                     // with wifi the service will be rescheduled on the interval.
                     // Wifi connection state changes all the time and starting services is expensive
                     // so it's important to only do this if we have stored events.
-                    ServiceScheduler.startService(context, EventIntentService.JOB_ID, eventServiceIntent);
-                    logger.info("Preemptively flushing events since wifi became available");
+                    if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
```
jophde: Why not schedule on Android O?

thomaszurkan-optimizely: I'm adding a comment.  But, it is because in Android O and higher, we use the job scheduler and make the job persistent.  this means that it will restart itself.

PR #276, opened 2019-06-26
(chore): Use the project config manager in java sdk with android. Also add listener.

thomaszurkan-optimizely:
## Summary
- This implements the new project config.  It also integrates with the android datafile handler polling.  By using a FileObserver available in Android, we update the project config when the cached file is modified.

What this means is that now when a new datafile comes in for android, the config is updated on the fly.
I haven't added any tests to this yet.  But, I wanted to get some comments more sooner than later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
 android:
   components:
     - tools
-    - platform-tools
+    - platform-tools-$ANDROID_BUILD_TOOLS
     - build-tools-$ANDROID_BUILD_TOOLS
     - android-$ANDROID_API
     - android-$EMULATOR_API
@@ -34,7 +34,7 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 before_install:
-    - yes | sdkmanager "platforms;android-$ANDROID_API"
+    #- yes | sdkmanager "platforms;android-$ANDROID_API"
 before_script:
   - echo $TRAVIS_BRANCH
   - echo $TRAVIS_TAG
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: android
 env:
  global:
 # These parameters should match the parameters for build tools and sdk versions in the gradle file
- - ANDROID_BUILD_TOOLS=28.0.3 # should match gradle
+ - ANDROID_BUILD_TOOLS=29.0.0 # should match gradle
  - ADB_INSTALL_TIMEOUT=5 # minutes
- - ANDROID_API=28 # api is same as gradle file
+ - ANDROID_API=29 # api is same as gradle file
  matrix:
  - EMULATOR_API=22
  - EMULATOR_API=21
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,8 @@ env:
  - ADB_INSTALL_TIMEOUT=5 # minutes
  - ANDROID_API=29 # api is same as gradle file
  matrix:
- - EMULATOR_API=22
- - EMULATOR_API=21
+ - EMULATOR_API=24
+ - EMULATOR_API=28
 android:
   components:
     - tools
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,12 @@ env:
  - ADB_INSTALL_TIMEOUT=5 # minutes
  - ANDROID_API=29 # api is same as gradle file
  matrix:
- - EMULATOR_API=24
- - EMULATOR_API=28
+ - EMULATOR_API=21
+ - EMULATOR_API=22
 android:
   components:
     - tools
     - platform-tools
-    - tools
     - build-tools-$ANDROID_BUILD_TOOLS
     - android-$ANDROID_API
     - android-$EMULATOR_API
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: android
 env:
  global:
 # These parameters should match the parameters for build tools and sdk versions in the gradle file
- - ANDROID_BUILD_TOOLS=29.0.0 # should match gradle
+ - ANDROID_BUILD_TOOLS=28.0.3 # should match gradle
  - ADB_INSTALL_TIMEOUT=5 # minutes
- - ANDROID_API=29 # api is same as gradle file
+ - ANDROID_API=28 # api is same as gradle file
  matrix:
  - EMULATOR_API=21
  - EMULATOR_API=22
@@ -34,7 +34,7 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 before_install:
-    - yes | sdkmanager "platforms;android-29"
+    - yes | sdkmanager "platforms;android-$ANDROID_API"
 before_script:
   - echo $TRAVIS_BRANCH
   - echo $TRAVIS_TAG
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_script:
   - echo $TRAVIS_BRANCH
   - echo $TRAVIS_TAG
   - echo no | android create avd --force -n test -t android-$EMULATOR_API --abi armeabi-v7a
-  - emulator -avd test -no-audio -no-window &
+  - emulator -avd test -no-window &
   - android-wait-for-emulator
   - adb shell input keyevent 82 &
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 before_install:
-    - yes | sdkmanager "platforms;android-28"
+    - yes | sdkmanager "platforms;android-29"
 before_script:
   - echo $TRAVIS_BRANCH
   - echo $TRAVIS_TAG
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_script:
   - echo $TRAVIS_BRANCH
   - echo $TRAVIS_TAG
   - echo no | android create avd --force -n test -t android-$EMULATOR_API --abi armeabi-v7a
-  - emulator -avd test -no-window &
+  - emulator -avd test -no-audio -no-window &
   - android-wait-for-emulator
   - adb shell input keyevent 82 &
 addons:
--- a/android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyClientTest.java
+++ b/android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyClientTest.java
@@ -356,7 +356,7 @@ public class OptimizelyClientTest {
         Experiment experiment = optimizelyClient.getProjectConfig().getExperimentKeyMapping().get(FEATURE_ANDROID_EXPERIMENT_KEY);
         attributes.put(BUCKETING_ATTRIBUTE, bucketingId);
         Variation v = optimizelyClient.activate(FEATURE_ANDROID_EXPERIMENT_KEY, GENERIC_USER_ID, attributes);
-        verify(bucketer).bucket( experiment, bucketingId, optimizely.getProjectConfig());
+        verify(bucketer).bucket(experiment, bucketingId, optimizely.getProjectConfig());
     }
 
     @Test
--- a/android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyClientTest.java
+++ b/android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyClientTest.java
@@ -115,9 +115,9 @@ public class OptimizelyClientTest {
             eventHandler = spy(DefaultEventHandler.getInstance(InstrumentationRegistry.getTargetContext()));
             optimizely = Optimizely.builder(datafile, eventHandler).build();
             if(datafileVersion==3) {
-                when(bucketer.bucket(optimizely.getProjectConfig().getExperiments().get(0), GENERIC_USER_ID)).thenReturn(optimizely.getProjectConfig().getExperiments().get(0).getVariations().get(0));
+                when(bucketer.bucket(optimizely.getProjectConfig().getExperiments().get(0), GENERIC_USER_ID, optimizely.getProjectConfig())).thenReturn(optimizely.getProjectConfig().getExperiments().get(0).getVariations().get(0));
             } else {
-                when(bucketer.bucket(optimizely.getProjectConfig().getExperimentKeyMapping().get(FEATURE_MULTI_VARIATE_EXPERIMENT_KEY), GENERIC_USER_ID)).thenReturn(optimizely.getProjectConfig().getExperimentKeyMapping().get(FEATURE_MULTI_VARIATE_EXPERIMENT_KEY).getVariations().get(1));
+                when(bucketer.bucket(optimizely.getProjectConfig().getExperimentKeyMapping().get(FEATURE_MULTI_VARIATE_EXPERIMENT_KEY), GENERIC_USER_ID, optimizely.getProjectConfig())).thenReturn(optimizely.getProjectConfig().getExperimentKeyMapping().get(FEATURE_MULTI_VARIATE_EXPERIMENT_KEY).getVariations().get(1));
             }
             spyOnConfig();
         } catch (Exception configException) {
@@ -356,7 +356,7 @@ public class OptimizelyClientTest {
         Experiment experiment = optimizelyClient.getProjectConfig().getExperimentKeyMapping().get(FEATURE_ANDROID_EXPERIMENT_KEY);
         attributes.put(BUCKETING_ATTRIBUTE, bucketingId);
         Variation v = optimizelyClient.activate(FEATURE_ANDROID_EXPERIMENT_KEY, GENERIC_USER_ID, attributes);
-        verify(bucketer).bucket( experiment, bucketingId);
+        verify(bucketer).bucket( experiment, bucketingId, optimizely.getProjectConfig());
     }
 
     @Test
@@ -881,7 +881,7 @@ public class OptimizelyClientTest {
         Map<String, String> attributes = new HashMap<>();
         attributes.put(BUCKETING_ATTRIBUTE, bucketingId);
         Variation v = optimizelyClient.getVariation("android_experiment_key", "userId", attributes);
-        verify(bucketer).bucket(experiment, bucketingId);
+        verify(bucketer).bucket(experiment, bucketingId, optimizely.getProjectConfig());
     }
 
     @Test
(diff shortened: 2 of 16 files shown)


datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -122,6 +144,28 @@ public void startBackgroundUpdates(Context context, DatafileConfig datafileConfi
...
         storeInterval(context, updateInterval * 1000);
+
+        DatafileCache datafileCache = new DatafileCache(
+                datafileConfig.getKey(),
+                new Cache(context, LoggerFactory.getLogger(Cache.class)),
+                LoggerFactory.getLogger(DatafileCache.class)
+        );
+
+        File filesFolder = context.getFilesDir();
+        fileObserver = new FileObserver(filesFolder.getPath()) {
```
mikecdavis: should we check that `fileObserver == null`? Just curious what would happen if `startBackgroundUpdates` was called multiple times?

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
```diff
@@ -78,6 +80,12 @@ protected void onStart() {
 
                    System.out.println("got track");
                });
+               optimizelyManager.getOptimizely().getNotificationCenter().addNotificationHandler(UpdateConfigNotification.class, new NotificationHandler<UpdateConfigNotification>() {
```
mikecdavis: Nit, can use a lambda.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -19,24 +19,46 @@
...
+import java.io.File;
+import java.util.function.Function;
+
 /**
  * The default implementation of {@link DatafileHandler} and the main
  * interaction point to the datafile-handler module.
  */
 public class DefaultDatafileHandler implements DatafileHandler {
+    private Logger logger = LoggerFactory.getLogger("DefaultDatafileHandler");
+    private DatafileConfig datafileConfig;
```
mikecdavis: This member variable is never accessed since a datafileConfig is passed in with each method call.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -19,24 +19,46 @@
...
+    private ProjectConfig currentProjectConfig;
     private DatafileServiceConnection datafileServiceConnection;
+    private FileObserver fileObserver;
+
+    public DefaultDatafileHandler(String sdkKey) {
+        this(new DatafileConfig(null, sdkKey));
+    }
+
+    public DefaultDatafileHandler(DatafileConfig datafileConfig) {
+        this.datafileConfig = datafileConfig;
```
mikecdavis: Not used.

.travis.yml
```diff
@@ -35,7 +35,7 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 before_install:
-    - yes | sdkmanager "platforms;android-28"
+    - yes | sdkmanager "platforms;android-29"
```
aliabbasrizvi: Isn't it possible to use `ANDROID_API` variable here?

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileChangeListener.java
```diff
@@ -0,0 +1,5 @@
+package com.optimizely.ab.android.datafile_handler;
+
+public interface DatafileChangeListener {
```
mikecdavis: Do we need this listener? Can't we leverage the existing `DatafileLoadedListener`

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
```diff
@@ -78,6 +80,12 @@ protected void onStart() {
...
+                   @Override
+                   public void handle(UpdateConfigNotification message) {
```
aliabbasrizvi: nit. `message` is not the right choice of variable name I feel. Perhaps `updateNotification`.

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
```diff
@@ -78,6 +80,12 @@ protected void onStart() {
...
+                        System.out.println("got datafile change ");
```
aliabbasrizvi: nit. Extraneous space at end.

android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyClientTest.java
```diff
@@ -356,7 +356,7 @@ public void testGoodActivationBucketingId() {
         Experiment experiment = optimizelyClient.getProjectConfig().getExperimentKeyMapping().get(FEATURE_ANDROID_EXPERIMENT_KEY);
         attributes.put(BUCKETING_ATTRIBUTE, bucketingId);
         Variation v = optimizelyClient.activate(FEATURE_ANDROID_EXPERIMENT_KEY, GENERIC_USER_ID, attributes);
-        verify(bucketer).bucket( experiment, bucketingId);
+        verify(bucketer).bucket( experiment, bucketingId, optimizely.getProjectConfig());
```
aliabbasrizvi: nit. Remove space between ( and experiment.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -242,4 +291,21 @@ public void removeSavedDatafile(Context context, DatafileConfig datafileConfig)
             datafileCache.delete();
         }
     }
+
+    public void setDatafile(String datafile) {
+        if (currentProjectConfig == null && datafile != null && !datafile.isEmpty()) {
```
mikecdavis: Q: We only want to set this once?

Style thing, but I prefer to fail fast and to group logical conditions. I think it makes the code easier to grok, log and requires less nesting which helps readability.

```
public void setDatafile(String datafile) {
    if (currentProjectConfig == null) {
        logger.info("Datafile already set, ignoring update.");
        return;
    } 
    
    if (datafile != null && !datafile.isEmpty()) {
        logger.info("datafile is empty, ignoring update");
        return;
    }
        
    try {
        currentProjectConfig = new DatafileProjectConfig.Builder().withDatafile(datafile).build();
        logger.info("Datafile successfully loaded with revision: {}", currentProjectConfig.getRevision());
    } catch (ConfigParseException ex) {
        logger.error("Unable to parse the datafile", ex);
        logger.info("Datafile is invalid: {}", datafile);
    }
}
```

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -498,6 +509,14 @@ private OptimizelyClient buildOptimizely(@NonNull Context context, @NonNull Stri
             builder.withUserProfileService(userProfileService);
         }
 
+        if (datafileHandler != null) {
```
mikecdavis: This is never null.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -498,6 +509,14 @@ private OptimizelyClient buildOptimizely(@NonNull Context context, @NonNull Stri
...
+            if (datafileHandler instanceof DefaultDatafileHandler) {
+                DefaultDatafileHandler handler = (DefaultDatafileHandler)datafileHandler;
+                handler.setDatafile(datafile);
```
mikecdavis: Shouldn't these defaults be part of the `OptimizelyManager.Builder` and not defaulted here?

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -19,24 +19,46 @@
...
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
+import java.util.function.Function;
+
 /**
  * The default implementation of {@link DatafileHandler} and the main
  * interaction point to the datafile-handler module.
  */
 public class DefaultDatafileHandler implements DatafileHandler {
```
aliabbasrizvi: Can we not remove reliance on `DefaultDatafileHandler` and move to an implementation of [`ProjectConfigManager`](https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/config/ProjectConfigManager.java)

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: there is a stop in the start.  the stop clears the fileObserver properly.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileChangeListener.java
thomaszurkan-optimizely: Didn't want to overload datafile load listener.  Also, the one method per interface allows for lambdas.

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
thomaszurkan-optimizely: this is the test-app and it was what the notification defaults to.  the change you are requesting would be at the java level.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: datafile handler can be instantiated by the use and passed into the constructor.  this is meant for that.  it is also used by the convenience constructor.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: my bad.  the currentProjectConfig is not checked.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: This is more for backward compatibility.  We also have issues where mobile developers have wanted more control.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: The question here, is should deprecate the old methods that passed in a project config and use the internal dataconfig created?

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileChangeListener.java
mikecdavis: Not sure it's overloading as semantically we are "loading" a datafile.
JavaDoc for DatafileLoadedListener:
> Listens for new Optimizely datafiles

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
mikecdavis: thanks. consider, removing the nesting by exiting early. Think of the empty check as a precondition.
```
if (datafile != null && !datafile.isEmpty()) {
    logger.info("datafile is empty, ignoring update");
    return;
}
```
this also adds visibility, etc.

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.java
mikecdavis: yeah `NotificationManager<T>` is a generic class with `handle(T message)` as the signature.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileHandler.java
```diff
@@ -18,14 +18,18 @@
...
 
+import java.util.function.Function;
+
 /**
  * DatafileHandler
  * class that is used to interact with the datafile_handler module. This interface can be
  * overridden so that the sdk user can provide a override for the default DatafileHandler.
  */
-public interface DatafileHandler {
+public interface DatafileHandler extends ProjectConfigManager {
```
mikecdavis: I think these changes break backwards compatibility with custom implementations.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: I'm not sure I follow you.  Are you saying that this check should be pulled out of the method?  In general I think that if you are taking a more functional approach (not easy in java), you want functions to be self contained and not be potentially fatal because of bad data.  Am I following this correctly?

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: I marked DatafileHandler interface as deprecated.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileHandler.java
```diff
@@ -18,13 +18,21 @@
...
+import com.optimizely.ab.config.ProjectConfig;
+import com.optimizely.ab.config.ProjectConfigManager;
 import com.optimizely.ab.android.shared.DatafileConfig;
 
+import java.util.function.Function;
+
 /**
  * DatafileHandler
+ * @deprecated
+ * This interface will be replaced by the ProjectConfigManager.  If you are implementing this interface moving forward,
```
mikecdavis: Can we `@link` to ProjectConfigManager?

android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyManagerTest.java
```diff
@@ -261,6 +272,7 @@ public void initializeWithEmptyDatafile() {
         Context appContext = mock(Context.class);
         when(context.getApplicationContext()).thenReturn(appContext);
         when(appContext.getPackageName()).thenReturn("com.optly");
+        when(defaultDatafileHandler.getConfig()).thenReturn(null);
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: the when override works fine.  these are unit tests.  however, it might also be possible to just include the default datafile handler instead of a mock and achieve the same thing.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -483,20 +488,26 @@ private OptimizelyClient buildOptimizely(@NonNull Context context, @NonNull Stri
 
         EventBatch.ClientEngine clientEngine = OptimizelyClientEngine.getClientEngineFromContext(context);
 
-        Optimizely.Builder builder = Optimizely.builder(datafile, eventHandler)
-                .withClientEngine(clientEngine)
+        Optimizely.Builder builder = Optimizely.builder();
+
+        if (datafileHandler instanceof DefaultDatafileHandler) {
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: At build time the developer has not called an initialize/start.  It was decided that there would be explicit ways to initialize sync and async.  So, moving it would break backward compatibility.  However, this does bring up the point on whether we should also be be deprecating the initialize methods.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -104,7 +109,7 @@ public void onStop(Context context) {
      * @param datafileConfig DatafileConfig for the datafile
      * @param updateInterval frequency of updates in seconds
      */
-    public void startBackgroundUpdates(Context context, DatafileConfig datafileConfig, Long updateInterval) {
+    public void startBackgroundUpdates(Context context, DatafileConfig datafileConfig, Long updateInterval, DatafileLoadedListener listener) {
         // if already running, stop it
         stopBackgroundUpdates(context, datafileConfig);
```
thomaszurkan-optimizely: stop background updates clears out the previous update if there was any.  It also stops the file observer and sets it to null.

android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyManagerBuilderTest.java
```diff
@@ -29,6 +29,7 @@
 import com.optimizely.ab.OptimizelyRuntimeException;
 import com.optimizely.ab.android.datafile_handler.DefaultDatafileHandler;
 import com.optimizely.ab.android.event_handler.DefaultEventHandler;
+import com.optimizely.ab.android.shared.DatafileConfig;
```
aliabbasrizvi: This import seems unnecessary.

android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyManagerTest.java
```diff
@@ -88,19 +91,24 @@
...
+        ProjectConfig config = new DatafileProjectConfig.Builder().withDatafile(datafile).build();
+
+        when(defaultDatafileHandler.getConfig()).thenReturn(config);
     }
 
     @Test
     public void initializeIntUseForcedVariation() {
         optimizelyManager.initialize(InstrumentationRegistry.getTargetContext(), R.raw.datafile);
 
+
```
aliabbasrizvi: Seems extraneous. Remove.

Issue #277, opened 2019-07-10
Unable to parse null datafile.


jia-tse-zocdoc:
Hi, 

We were previously on an older SDK `2.1.0` and made the jump to `3.0.0` and started noticing these kinds of error show up in our logs. Here is an example of a stacktrace.  It's not reproducible locally, and only affects a small percentage of our users. Not sure what's causing it, but we believe the issue is not related to internet connectivity. 

```
Caused by com.optimizely.ab.config.parser.ConfigParseException: Unable to parse null datafile.
       at com.optimizely.ab.config.ProjectConfig$Builder.build + 605(ProjectConfig.java:605)
       at com.optimizely.ab.Optimizely.initialize + 119(Optimizely.java:119)
       at com.optimizely.ab.Optimizely$Builder.build + 941(Optimizely.java:941)
       at com.optimizely.ab.android.sdk.OptimizelyManager.buildOptimizely + 501(OptimizelyManager.java:501)
       at com.optimizely.ab.android.sdk.OptimizelyManager.injectOptimizely + 445(OptimizelyManager.java:445)
       at com.optimizely.ab.android.sdk.OptimizelyManager$2.onDatafileLoaded + 343(OptimizelyManager.java:343)
       at com.optimizely.ab.android.datafile_handler.DefaultDatafileHandler$1.onDatafileLoaded + 78(DefaultDatafileHandler.java:78)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader.notify + 81(DatafileLoader.java:81)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader.access$000 + 35(DatafileLoader.java:35)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader$RequestDatafileFromClientTask.onPostExecute + 134(DatafileLoader.java:134)
       at com.optimizely.ab.android.datafile_handler.DatafileLoader$RequestDatafileFromClientTask.onPostExecute + 85(DatafileLoader.java:85)
       at android.os.AsyncTask.finish + 695(AsyncTask.java:695)
       at android.os.AsyncTask.-wrap1(AsyncTask.java)
       at android.os.AsyncTask$InternalHandler.handleMessage + 712(AsyncTask.java:712)
       at android.os.Handler.dispatchMessage + 106(Handler.java:106)
       at android.os.Looper.loop + 164(Looper.java:164)
       at android.app.ActivityThread.main + 6626(ActivityThread.java:6626)
[...]

PR #278, opened 2019-07-10
(fix): a fix for problem where the datafile cache has been corrupted or remo…

thomaszurkan-optimizely:
…ved.

## Summary
- If the cached datafile is removed or corrupted, and last modified has already been set, the app won't load.
- If loading synchronously there is a similar problem where the datafile cache exists but is corrupt, it will also fail to initialize.
## Issues
- #277

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
@@ -279,8 +279,13 @@ public class OptimizelyManager {
     public String getDatafile(Context context,@RawRes Integer datafileRes){
         try {
             if (isDatafileCached(context)) {
-                return datafileHandler.loadSavedDatafile(context, datafileConfig);
-            } else if (datafileRes!=null) {
+                String datafile = datafileHandler.loadSavedDatafile(context, datafileConfig);
+                if (datafile != null) {
+                    return datafile;
+                }
+            }
+
+            if (datafileRes!=null) {
                 return loadRawResource(context, datafileRes);
             }else{
                 logger.error("Invalid datafile resource ID.");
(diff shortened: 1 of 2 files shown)


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -279,8 +279,13 @@ public void run() {
...
             if (isDatafileCached(context)) {
-                return datafileHandler.loadSavedDatafile(context, datafileConfig);
-            } else if (datafileRes!=null) {
+                String datafile = datafileHandler.loadSavedDatafile(context, datafileConfig);
+                if (datafile != null) {
+                    return datafile;
+                }
+            }
+
+            if (datafileRes!=null) {
```
mikeng13: nit: spacing before and after the `!=`

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileLoader.java
```diff
@@ -110,6 +111,12 @@ private void notify(@Nullable DatafileLoadedListener datafileLoadedListener, @Nu
 
         @Override
         protected String doInBackground(Void... params) {
+
+            if (!datafileCache.exists() || (datafileCache.exists() && datafileCache.load() == null)) {
+                // there is a problem with the cached datafile.  set last modified to 1970
```
mikeng13: This comment doesn't really make sense about the code below. Why are we creating an new storage instance here?

thomaszurkan-optimizely: It's a wrapper around the context storage.  I am creating another wrapper.

PR #280, opened 2019-07-11
(fix): Corrupt datafile cache fix

thomaszurkan-optimizely:
## Summary
- If the cached datafile is removed or corrupted, and last modified has already been set, the app won't load.
- If loading synchronously there is a similar problem where the datafile cache exists but is corrupt, it will also fail to initialize.
## Issues
- #277


datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileLoader.java
```diff
@@ -110,6 +111,14 @@ private void notify(@Nullable DatafileLoadedListener datafileLoadedListener, @Nu
 
         @Override
         protected String doInBackground(Void... params) {
+
+            // if there is a problem with the cached datafile, set last modified to 1970
+            if (!datafileCache.exists() || (datafileCache.exists() && datafileCache.load() == null)) {
+                // create a wrapper for application context default storage.
+                OptlyStorage storage = new OptlyStorage(this.datafileService.getApplicationContext());
+                // set the last modified for this url to 1 millisecond passed Jan 1, 1970.
```
mikeng13: nit: I think it's "past" and not "passed"

PR #299, opened 2019-08-22
(fix): Refactored Use of Event Processor

fayyazarshad:
## Summary
- Added EventProcessor in the parameters list for creating Optimizely Instance.
- Refactored Optimizely Test by removing unused imports
- Fixed test by adding Event Processor in params

## Test plan
Fixed tests in `OptimizelyManagerTest.java`

Note: Currently Unit Tests are being failed as the version Android SDK is using in this branch doesn't have code for Batch Event Processor in Java SDK. But I have tested it by changing the release to 3.4.0-beta of Java SDK and these tests were passed, then I reverted the release to 3.2.1.

### DIFF
--- a/android-sdk/build.gradle
+++ b/android-sdk/build.gradle
@@ -62,7 +62,7 @@ dependencies {
     api project(':datafile-handler')
     api project(':event-handler')
     api project(':user-profile')
-    api ("com.optimizely.ab:core-api:$java_core_ver") {
+    api ("com.optimizely.ab:core-api:3.4.0-beta") {
         exclude group: 'com.google.code.findbugs'
     }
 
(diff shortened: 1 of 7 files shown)


fayyazarshad:
@mikecdavis @thomaszurkan-optimizely This PR is available for review.

Note: Currently Unit Tests are being failed as the version Android SDK is using in this branch doesn't have code for Batch Event Processor in Java SDK. But I have tested it by changing the release to 3.4.0-beta of Java SDK and these tests were passed, then I reverted the release to 3.2.1.

mikeng13:
@mnoman09 are the unit test failures legitimate? Can you look into it? Also, can you a version of FSC with event batching enabled for this branch and link it here please?

PR #303, opened 2019-09-10
(fix): speed up the initialize by priming the config parser.

thomaszurkan-optimizely:
## Summary
- this gives a performance boost by priming the config parser and cleanup some code.

## Test
Augmented the test-app so that it is paused on the splash screen waiting for a button click before initializing the SDK.  This allows us to use Android Studio profiling to find method times.

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
@@ -108,9 +108,9 @@ public class OptimizelyManager {
         this.errorHandler = errorHandler;
         this.userProfileService = userProfileService;
 
-        // prime the parser
-        // so it isn't included in initialization cost.
+        // prime the config parser if it is lazy loaded.
         DefaultConfigParser.getInstance();
+
     }
 
     @VisibleForTesting
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
@@ -40,7 +40,9 @@ import com.optimizely.ab.android.shared.DatafileConfig;
 import com.optimizely.ab.android.user_profile.DefaultUserProfileService;
 import com.optimizely.ab.bucketing.UserProfileService;
 import com.optimizely.ab.config.ProjectConfig;
+import com.optimizely.ab.config.ProjectConfigManager;
 import com.optimizely.ab.config.parser.ConfigParseException;
+import com.optimizely.ab.config.parser.DefaultConfigParser;
 import com.optimizely.ab.error.ErrorHandler;
 import com.optimizely.ab.event.EventHandler;
 import com.optimizely.ab.event.internal.payload.EventBatch;
@@ -105,6 +107,10 @@ public class OptimizelyManager {
         this.eventHandler = eventHandler;
         this.errorHandler = errorHandler;
         this.userProfileService = userProfileService;
+
+        // prime the parser
+        // so it isn't included in initialization cost.
+        DefaultConfigParser.getInstance();
     }
 
     @VisibleForTesting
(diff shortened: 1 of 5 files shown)


datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -291,17 +293,38 @@ public void setDatafile(String datafile) {
...
-            logger.info("Datafile is invalid");
+    private ProjectConfig safeGetConfig() {
+        synchronized (projectConfigLock) {
+            return currentProjectConfig;
         }
     }
     @Override
     public ProjectConfig getConfig() {
-        return currentProjectConfig;
+        return safeGetConfig();
```
mikecdavis: Since we're now loading the config asynchronously, should we block here until the config is set? We might be protecting against this somewhere else, I'm just not sure if this changes the initialization semantics in a backwards incompatible way. With change this can return `null`, while we're loading and parsing the cached datafile.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -291,17 +293,38 @@ public void setDatafile(String datafile) {
...
+                return null;
+            }
+        };
+
+        task.execute();
+
+    }
+
+    private void safeSetConfig(String datafile) {
+        synchronized (projectConfigLock) {
```
mikecdavis: We're locking access to the ProjectConfig while deserializing. This unnecessarily blocks during this computation step from a background thread. I think using `AsyncTask#doInBackground` and `AsyncTask#onPostExecute` as described above would eliminate the need for any locking since variable assignments are atomic.
- http://etutorials.org/Programming/Java+performance+tuning/Chapter+10.+Threading/10.6+Atomic+Access+and+Assignment/

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -291,17 +293,38 @@ public void setDatafile(String datafile) {
             return;
         }
 
-        try {
-            currentProjectConfig = new DatafileProjectConfig.Builder().withDatafile(datafile).build();
+        AsyncTask task = new AsyncTask() {
```
mikecdavis: I'm new to `AsyncTask`, but one source I read said "Operations in this method should not touch on any mainthread activities or fragments." when referring to `doInBackground`
- https://www.tutorialspoint.com/android-asynctask-example-and-explanation

Instead they recommend using `onPostExecute(Result)` to update the main thread, so this would change to be something like:
```
AsyncTask<String, Object, ProjectConfig> task = new AsyncTask<>() {
    @Override
    protected ProjectConfig doInBackground(String[] datafiles) {
        try {
            return new DatafileProjectConfig.Builder().withDatafile(datafile).build();
        } catch (ConfigParseException ex) {
            logger.error("Unable to parse the datafile", ex);
        }
        return null;
    }

    @Override
    protected void onPostExecute(ProjectConfig projectConfig) {
        if (projectConfig == null) {
            logger.info("Datafile is invalid");
            return;
        }

        logger.info("Datafile successfully loaded with revision: {}", currentProjectConfig.getRevision());
        currentProjectConfig = projectConfig;
    }

task.execute(datafile);
```

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -105,6 +107,10 @@
         this.eventHandler = eventHandler;
         this.errorHandler = errorHandler;
         this.userProfileService = userProfileService;
+
+        // prime the parser
+        // so it isn't included in initialization cost.
+        DefaultConfigParser.getInstance();
```
mikecdavis: Can you elaborate on how this benefits initialization cost? Since we're also moving all dependent operations into background tasks doesn't this move any reflection cost back into the main thread?

PR #313, opened 2020-01-07
fix (proguard): moved slf4j proguard keep to sdk rules

thomaszurkan-optimizely:
## Summary
- Since it is required in a minify.  We can probably add this rule without too much complaint.

## Test plan
- I generated a signed release pack with minify to insure that everything still works


thomaszurkan-optimizely:
@aliabbasrizvi it is in 3.1.2.  It is not in 3.3.x yet.

PR #318, opened 2020-02-20
fix: change sync-init flow to support project auto update on cache download

jaeopt:
## Summary
- Synchronous init has one more config parameter (updateConfigOnNewDatafile).
- When it's set true, the project config is automatically updated when an updated datafile is downloaded for caching. Otherwise, the cached datafile is used when the SDK is started next time (default).

## Issues
- OASIS-5957


datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -128,6 +137,19 @@ public void startBackgroundUpdates(Context context, DatafileConfig datafileConfi
 
         storeInterval(context, updateInterval * 1000);
 
+        enableUpdateConfigOnNewDatafile(context, datafileConfig, listener);
+    }
+
+    public void enableUpdateConfigOnNewDatafile(Context context, DatafileConfig datafileConfig, DatafileLoadedListener listener) {
+        if (listener != null) {
```
thomaszurkan-optimizely: i don't think the listener should exist here.  it should be removed.  also, if you are going to store the listener as property, you need to clear it out along with the file observer when stop is called.

PR #319, opened 2020-02-24
chore: clarify min polling interval as 15 mins

jaeopt:
## Summary
- Clarify minimum polling interval as 15 mins
- Fix the sample code accordingly

OASIS-5988


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -677,6 +677,7 @@ public void onActivityDestroyed(Activity activity) {
          * Sets the interval which {@link DatafileService} through the {@link DatafileHandler} will attempt to update the
          * cached datafile.  If you set this to -1, you disable background updates.  If you don't set
          * a download interval (or set to less than 0), then no background updates will be scheduled or occur.
+         * The minimum interval is 900 secs (15 minutes, enforced by the Android JobScheduler API)
```
thomaszurkan-optimizely: nit: can we add a link to the job scheduler java doc? also, mention what API version that is associated with.

Issue #320, opened 2020-02-25
Main Thread IO done on SDK initialization


MihaiHoinarescu:
Everytime the SDK initializes we get a bunch of strict mode violations.

It seems like the SDK is doing some IO in onPostExecute of an AsyncTask. It would be great if this IO could be done on a different thread so it doesn't slow down the user experience.

I've added the stack trace below.

 ```01-21 19:15:06.023 18420 18420 D StrictMode: StrictMode policy violation; ~duration=106 ms: android.os.strictmode.DiskReadViolation
01-21 19:15:06.023 18420 18420 D StrictMode: at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1556)
01-21 19:15:06.023 18420 18420 D StrictMode: at libcore.io.BlockGuardOs.access(BlockGuardOs.java:69)
01-21 19:15:06.023 18420 18420 D StrictMode: at libcore.io.ForwardingOs.access(ForwardingOs.java:73)
01-21 19:15:06.023 18420 18420 D StrictMode: at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:7246)
01-21 19:15:06.023 18420 18420 D StrictMode: at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:281)
01-21 19:15:06.023 18420 18420 D StrictMode: at java.io.File.exists(File.java:815)
01-21 19:15:06.023 18420 18420 D StrictMode: at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:645)
01-21 19:15:06.023 18420 18420 D StrictMode: at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:636)
01-21 19:15:06.023 18420 18420 D StrictMode: at android.app.ContextImpl.getFilesDir(ContextImpl.java:681)
01-21 19:15:06.023 18420 18420 D StrictMode: at android.app.ContextImpl.fileList(ContextImpl.java:792)
01-21 19:15:06.023 18420 18420 D StrictMode: at android.content.ContextWrapper.fileList(ContextWrapper.java:233)
01-21 19:15:06.023 18420 18420 D StrictMode: at com.optimizely.ab.android.shared.Cache.exists(Cache.java:69)
01-21 19:15:06.023 18420 18420 D StrictMode: at com.optimizely.ab.android.datafile_handler.DatafileCache.exists(DatafileCache.java:65)
[...]

PR #321, opened 2020-02-27
(fix): Listener on task

thomaszurkan-optimizely:
## Summary
- Load the cached datafile in the datafile loader if the datafile is null from cdn.  This avoids loading the cached datafile in the onPostExecute of the datafile loader which is on the main thread.

## Issues
- #320

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
@@ -21,6 +21,7 @@ import android.app.Activity;
 import android.app.Application;
 import android.content.Context;
 import android.content.res.Resources;
+import android.os.AsyncTask;
 import android.os.Build;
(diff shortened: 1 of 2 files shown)


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -285,12 +285,12 @@ public void run() {
...
-                if (datafile != null) {
-                    return datafile;
-                }
-            }
+//            if (isDatafileCached(context)) {
+//                String datafile = datafileHandler.loadSavedDatafile(context, datafileConfig);
+//                if (datafile != null) {
+//                    return datafile;
+//                }
+//            }
```
jaeopt: It looks like this function is shared for sync and async init. I see you move this to async download thread. It looks great.
This will change sync init behavior (it'll skip cache load) if we remove it. We probably need a separate path for sync and async for getDatafile().

PR #328, opened 2020-04-06
chore: migrate android.support libraries to androidx

jaeopt:
## Summary
- Migrate android.support libraries to androidx


yogurtearl:
@jaeopt do you have plans to get this merged soon?
do you want any assistance with this androidx PR?

jaeopt:
@yogurtearl Thanks for the offer. We're holding it for applications still using the legacy support libraries. Do you see any issues using the SDK and need us to migrate to androidx?

yogurtearl:
@jaeopt we need it to be upgraded so that we can turn off jetifier in our app, which slows down our build times.

If an app hasn't upgraded to AndroidX in the last 2 years, they probably will stick with the old version of the optimizely sdk as well. :) 
Legacy apps can still continue to use your previously published SDKs that use the deprecated dependencies.

jaeopt:
@yogurtearl It's a good point. Thanks!

yogurtearl:
@jaeopt so... can we get this merged for an upcoming release? 😬

jaeopt:
@yogurtearl Cannot promise that but we'll look into it with your input.

nelljohan:
Another option here could be to publish a `-jetified` and a legacy dependency so that developers can have the choice 😀

This is the approach other libraries like https://github.com/Appboy/appboy-android-sdk have taken.

technoir42:
Please consider this for the next release. Optimizely is literally the last library out of several dozens in our project which hasn't been migrated to AndroidX yet.  
We, users, just want to turn off Jetifier and enjoy improved build times. Why harm the experience of the vast majority because of a few users which haven't migrated from the support library? They can continue using the old version of the library, or if they absolutely need new features they can [de-jetify](https://ncorti.com/blog/jetifier-reverse) the SDK.

jaeopt:
@technoir42 Thanks for your input. A good point. It'll be considered.

kozaxinan:
Version3.7 is released without AndroidX support. Optimizely one of last two libraries to drop Jetifier. Is there a timeline for this PR?

PR #330, opened 2020-05-14
fix: add notification convenience apis with sample codes

jaeopt:
## Summary
- Add convenience apis for UpdateConfigNotification and LogEventNotification to be consistent with other notifications
- Cleaned up sample codes for notifications both in sync and async init modes

## Issues
- OASIS-6085


android-sdk/src/androidTest/java/com/optimizely/ab/android/sdk/OptimizelyClientTest.java
```diff
@@ -1860,4 +1861,59 @@ public void handle(DecisionNotification decisionNotification) {
...
         assertFalse(optimizelyClient.getNotificationCenter().removeNotificationListener(notificationId2));
     }
+
+    @Test
+    public void testAddUpdateConfigNotificationHandler() {
+        OptimizelyClient optimizelyClient = new OptimizelyClient(
+                optimizely,
+                logger
+        );
+        NotificationManager<TrackNotification> manager = optimizely.getNotificationCenter()
```
thomaszurkan-optimizely, line 1871: This should be inline instead of getting the manager first.

test-app/src/main/java/com/optimizely/ab/android/test_app/SplashScreenActivity.kt
```diff
@@ -44,16 +43,53 @@ class SplashScreenActivity : AppCompatActivity() {
...
+
         // with the new Android O differences, you need to register the service for the intent filter you desire in code instead of
         // in the manifest.
         val eventRescheduler = EventRescheduler()
         applicationContext.registerReceiver(eventRescheduler, IntentFilter(WifiManager.SUPPLICANT_CONNECTION_CHANGE_ACTION))
-        myApplication?.let {
-            optimizelyManager?.initialize(it, null) { _ ->
+
+        if (INITIALIZE_ASYNCHRONOUSLY) {
+            optimizelyManager!!.initialize(this, R.raw.datafile) { _ ->
```
thomaszurkan-optimizely, line 54: don't you think it is unwise to use double bang?  it is supposed to be a example and i would never do that in prod code.

PR #335, opened 2020-06-12
fix: update to latest build tools. slight refactor of datafile service.

thomaszurkan-optimizely:
## Summary
- Eliminate piling up of datafile calls.  Use latest tools. 

## Test plan
- Lots of manual testing including taking the device offline for long periods of time.

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
@@ -228,7 +228,7 @@ public class OptimizelyManager {
             logger.error("Unable to build OptimizelyClient instance", e);
         }
 
-        if (downloadToCache) {
+        if (downloadToCache && !datafileDownloadEnabled()) {
             datafileHandler.downloadDatafileToCache(context, datafileConfig, updateConfigOnNewDatafile);
         }
 
@@ -491,8 +491,12 @@ public class OptimizelyManager {
         return datafileHandler;
     }
 
+    private boolean datafileDownloadEnabled() {
+        return datafileDownloadInterval > 0;
+    }
+
     private void startDatafileHandler(Context context) {
-        if (datafileDownloadInterval <= 0) {
+        if (!datafileDownloadEnabled()) {
             logger.debug("Invalid download interval, ignoring background updates.");
             return;
         }
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
@@ -228,7 +228,7 @@ public class OptimizelyManager {
             logger.error("Unable to build OptimizelyClient instance", e);
         }
 
-        if (downloadToCache && !datafileDownloadEnabled()) {
+        if (downloadToCache) {
             datafileHandler.downloadDatafileToCache(context, datafileConfig, updateConfigOnNewDatafile);
         }
 
--- a/datafile-handler/src/androidTest/java/com/optimizely/ab/android/datafile_handler/DatafileLoaderTest.java
+++ b/datafile-handler/src/androidTest/java/com/optimizely/ab/android/datafile_handler/DatafileLoaderTest.java
@@ -150,7 +150,7 @@ public class DatafileLoaderTest {
     public void warningsAreLogged() throws IOException {
         final ListeningExecutorService executor = MoreExecutors.newDirectExecutorService();
         Cache cache = mock(Cache.class);
-        datafileCache = new DatafileCache("1", cache, logger);
+        datafileCache = new DatafileCache("warningsAreLogged", cache, logger);
         DatafileLoader datafileLoader =
                 new DatafileLoader(datafileService, datafileClient, datafileCache, executor, logger);
 
@@ -159,7 +159,7 @@ public class DatafileLoaderTest {
         when(cache.delete(datafileCache.getFileName())).thenReturn(false);
         when(cache.save(datafileCache.getFileName(), "{}")).thenReturn(false);
 
-        datafileLoader.getDatafile("1", datafileLoadedListener);
+        datafileLoader.getDatafile("warningsAreLogged", datafileLoadedListener);
         try {
             executor.awaitTermination(5, TimeUnit.SECONDS);
         } catch (InterruptedException e) {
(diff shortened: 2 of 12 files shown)


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
```diff
@@ -228,7 +228,7 @@ public OptimizelyClient initialize(@NonNull Context context, @Nullable String da
             logger.error("Unable to build OptimizelyClient instance", e);
         }
 
-        if (downloadToCache) {
+        if (downloadToCache && !datafileDownloadEnabled()) {
```
jaeopt, line 231: Can we remove this extra flag check "datafileDownloadEnabled"?  
Clients still want to see datafile immediately updated even when they turn on polling interval (15min+). 
I see initial background fetch and periodic polling won't be called at the same time because of the big interval.

I think you found why we have a potential conflict in "startService" and setting the minLetency to one minute will fix it.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileRescheduler.java
```diff
@@ -95,11 +96,11 @@ void dispatch(Intent intent) {
             for (DatafileConfig datafileConfig : datafileConfigs) {
                 // for scheduled jobs Android O and above, we use the JobScheduler and persistent periodic jobs
                 // so, we don't need to do anything.
-               // if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
```
jaeopt, line 98: Remove this comment line

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
```diff
@@ -139,7 +140,7 @@ public void startBackgroundUpdates(Context context, DatafileConfig datafileConfi
         enableUpdateConfigOnNewDatafile(context, datafileConfig, listener);
     }
 
-    public void enableUpdateConfigOnNewDatafile(Context context, DatafileConfig datafileConfig, DatafileLoadedListener listener) {
+    synchronized public void enableUpdateConfigOnNewDatafile(Context context, DatafileConfig datafileConfig, DatafileLoadedListener listener) {
```
jaeopt, line 143: This sync looks good. 
For safety, let's add a synchronizer for event handler as well (line 160-173).

shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
```diff
@@ -109,7 +108,11 @@ private void setRepeating(long interval, PendingIntent pendingIntent, Intent int
                             ScheduledJobService.class.getName()));
             builder.setPeriodic(interval, interval);
             builder.setPersisted(true);
-            builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
```
jaeopt, line 112: I think this is the one you mentioned in the slack - removed to avoid accumulated calls when network is back up.
Is this a bug in new Android OS or a new feature?

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
thomaszurkan-optimizely: No.  This causes a double download.  It makes sense to have this here.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
thomaszurkan-optimizely: No

shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
thomaszurkan-optimizely: It has an unwanted side effect of just queuing up jobs rather than not running them. So, when a connection becomes available, it runs all queued.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileRescheduler.java
thomaszurkan-optimizely: that's what the red means.

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
jaeopt: I observed the double downloads. It looks like un-deterministic behavior of JobScheduler - polling download can be triggered any time between 0..15mins and in simulator, it's fired instantly (I think this will behave differently in real devices). 

This code fix will remove the double download issue when app starts. 
But in real devices, if clients expect immediate download with sync-initialization, it may behave differently (like 15min delayed) and can bring back the initialization issue we fixed before.

Can we find alt solution still supporting instant download even when polling enabled?

shared/src/main/java/com/optimizely/ab/android/shared/ServiceScheduler.java
```diff
@@ -109,7 +108,11 @@ private void setRepeating(long interval, PendingIntent pendingIntent, Intent int
...
+            // we are only doing repeating on datafile service. it is a prefetch.
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                builder.setPrefetch(true);
+            }
+            //builder.setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY);
```
mnoman09, line 115: Nit: remove?
Is there any reason to remove this line?

android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java
msohailhussain: I think I don't see any issue here. Agree with Tom.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DefaultDatafileHandler.java
msohailhussain: It's good to have sync on event handler but we already have sync on this method and disable. So chances for cross-thread is very low on event handler.

datafile-handler/src/androidTest/java/com/optimizely/ab/android/datafile_handler/DatafileLoaderTest.java
```diff
@@ -170,4 +170,29 @@ public void warningsAreLogged() throws IOException {
...
+        datafileLoader.getDatafile("debugLogged", datafileLoadedListener);
+        datafileLoader.getDatafile("debugLogged", datafileLoadedListener);
+        try {
+            executor.awaitTermination(5, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            fail();
+        }
+
+        verify(logger).debug("Last download happened under 1 minute ago. Throttled to be at least 1 minute apart.");
+        verify(datafileLoadedListener, atMost(2)).onDatafileLoaded("{}");
```
jaeopt, line 196: Let's also check if it's called at least once
```suggestion
        verify(datafileLoadedListener, atMost(2)).onDatafileLoaded("{}");
        verify(datafileLoadedListener, atLeast(1)).onDatafileLoaded("{}");
```

datafile-handler/src/androidTest/java/com/optimizely/ab/android/datafile_handler/DatafileLoaderTest.java
```diff
@@ -170,4 +170,29 @@ public void warningsAreLogged() throws IOException {
         verify(logger).warn("Unable to save new datafile");
         verify(datafileLoadedListener, atMost(1)).onDatafileLoaded("{}");
     }
+
+    @Test
+    public void debugLogged() throws IOException {
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: I don't want to add another an extra minute to the tests to prove this point.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileLoader.java
```diff
@@ -71,8 +101,11 @@ public void getDatafile(@NonNull String datafileUrl, @Nullable DatafileLoadedLis
                         this,
                         datafileLoadedListener,
                         logger);
+
         // Execute tasks in order
         requestDatafileFromClientTask.executeOnExecutor(executor);
+        // set the download time and don't allow downloads to overlap less than a minute
+        saveDownloadTime(datafileUrl);
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: I really don't see a reason to move this.  Can you explain?

datafile-handler/src/androidTest/java/com/optimizely/ab/android/datafile_handler/DatafileLoaderTest.java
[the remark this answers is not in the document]
jaeopt: Yep it's not great to add a minute wait time for testing. But this test must be important to make sure we do not freeze polling permanently by error. We can change it to inject the required gap time so can test with a reasonable wait time of a few seconds. 
I'm ok to release it as is for now and then add the extra test with the next coming release.

PR #348, opened 2020-09-29
Fix: Added slf4j as compileOnly

mnoman09:
## Summary
- Added slf4j as compileOnly 

## Test plan
All Unit tests and FSC should pass


aliabbasrizvi (approved):
@thomaszurkan-optimizely or @mikecdavis ?

thomaszurkan-optimizely (changes requested):
We don't add the slf4j android.  We use novoid because it supports the android logging properties file implemented in the test-app.  So, I think we should close this app.  That said, I know that with the update to the latest java-sdk we will still need the compile.  Thanks.

PR #358, opened 2021-01-25
fix: remove erroneous error if project id and sdk key are null.

### LINKED ISSUE #355: OptimizelyManager Builder requires projectId or sdkKey
Right now the `OptimizelyManager.Builder` requires either a `projectId` or a `sdkKey`. If neither are provided, `build()` will return [null](https://github.com/optimizely/android-sdk/blob/master/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyManager.java#L927-L930). One of these are required so that if a DatafileConfig is not provided, one can be created. However, if a DatafileConfig is provided, then neither projectId or sdkKey should be necessary. 
For instance, if one was to create an OptimizelyManager as such, then sdkKey or projectId should not be required.
```
val sdkKey = "myKey"
val datafileConfig = DatafileConfig(null, sdkKey, OPTIMIZELY_DATAFILE_HOST)
val optimizelyManager = OptimizelyManager.builder()
  .withDatafileConfig(datafileConfig)
  .withSDKKey(sdkKey) // not needed since we provide a DatafileConfig, but if not provided, build() returns null
  .build(context)
```

thomaszurkan-optimizely:
## Summary
-  move check to datafileConfig check.
-  simple change.
## Tests
- Added unit tests to builder test.
## Issues
- "Fixes #355 "


jaeopt (approved):
LGTM

PR #367, opened 2021-03-05
Fix: Removed isValidCondition from CreateUserContext

mnoman09:
## Summary
- Removed IsValid condition before creating OptimizelyUserContext
- This is to make sure the behavior is same like java sdk

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
@@ -794,12 +794,7 @@ public class OptimizelyClient {
     @Nullable
     public OptimizelyUserContext createUserContext(@NonNull String userId,
                                                    @NonNull Map<String, Object> attributes) {
-        if (isValid()) {
-            return optimizely.createUserContext(userId, attributes);
-        } else {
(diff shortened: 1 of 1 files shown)


android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
```diff
@@ -794,12 +794,7 @@ public OptimizelyConfig getOptimizelyConfig() {
     @Nullable
     public OptimizelyUserContext createUserContext(@NonNull String userId,
                                                    @NonNull Map<String, Object> attributes) {
-        if (isValid()) {
-            return optimizely.createUserContext(userId, attributes);
-        } else {
-            logger.warn("Optimizely is not initialized, could not create a user context");
-            return null;
-        }
+        return optimizely.createUserContext(userId, attributes);
```
jaeopt, line 797:
```suggestion
if (optimizely != null) {
            return optimizely.createUserContext(userId, attributes);
        } else {
            logger.warn("Optimizely is not initialized, could not create a user context");
            return null;
        }
```
It's a good catch - createContext should be allowed even when optimizely is not fully ready. But we still need null checking.

jaeopt (changes requested):
Can we add null checking?

jaeopt (approved):
LGTM

Issue #366, opened 2021-03-02
Datafile download fails on API 16


nathan-l:
Hi. 
When downloading a file with the default DataFileClient, `urlConnection.connect();` fails with an ssl error, here is the full stacktrace. 

```javax.net.ssl.SSLHandshakeException: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0x68a5f008: Failure in SSL library, usually a protocol error
error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure (external/openssl/ssl/s23_clnt.c:741 0x671c3d74:0x00000000)
	at com.android.org.conscrypt.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:448)
	at com.android.okhttp.Connection.upgradeToTls(Connection.java:146)
	at com.android.okhttp.Connection.connect(Connection.java:107)
	at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:294)
	at com.android.okhttp.internal.http.HttpEngine.sendSocketRequest(HttpEngine.java:255)
	at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:206)
	at com.android.okhttp.internal.http.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:345)
	at com.android.okhttp.internal.http.HttpURLConnectionImpl.connect(HttpURLConnectionImpl.java:89)
	at com.android.okhttp.internal.http.HttpsURLConnectionImpl.connect(HttpsURLConnectionImpl.java:161)
	at com.optimizely.ab.android.datafile_handler.DatafileClient$1.execute(DatafileClient.java:84)
	at com.optimizely.ab.android.datafile_handler.DatafileClient$1.execute(DatafileClient.java:68)
	at com.optimizely.ab.android.shared.Client.execute(Client.java:143)
	at com.optimizely.ab.android.datafile_handler.DatafileClient.request(DatafileClient.java:113)
	at com.optimizely.ab.android.datafile_handler.DatafileLoader$1.run(DatafileLoader.java:100)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
	at java.lang.Thread.run(Thread.java:841)
[...]

PR #373, opened 2021-05-05
fix: support datafile download and events upload for API-16

jaeopt:
## Summary
Android devices with API-19 and lower is set by default to use SSL lower than the our server requirements (TLS1.2+).
With this fix, those low-API devices will be set to use TLS1.2 for both datafile download and event upload.

## Test plan
- Unit tests for emulators with API-16 and API-19

## Issues
- OASIS-7555
- https://github.com/optimizely/android-sdk/issues/366


mnoman09 (changes requested):
overall looks good to me. Just one comment and all headers needs to get updated.

mnoman09 (approved):
LGTM

PR #383, opened 2021-08-31
fix(proguard): fix gson proguard rules and clean up other rules

jaeopt:
## Summary
- Fix proguard rules for GSON to one recommended by the GSON repo (https://github.com/google/gson/blob/master/examples/android-proguard-example/proguard.cfg)
- Clean up other proguard files.

## Test plan
- Build SNAPSHOT to validate obfuscation.
## Issues
- OASIS-7921


msohailhussain (approved):
checked with noman and lgtm

PR #386, opened 2021-09-28
fix: support arbitrarily long list of user attributes in batch events

jaeopt:
## Summary
- WorkManager has 10KB limit on Data, so we cannot support batch events for a long list of attributes or large batch sizes.
- BatchEvents are compressed/decompressed to support large events.
- Compression/decompression is bypassed when the data size is less than the 10KB limit.

## Test plan
- Add tests for compress/decompress for a range of batch event sizes.
- Test for backward compatibility for old OS versions.

## Issues
- "OASIS-8000"

### DIFF
--- a/event-handler/build.gradle
+++ b/event-handler/build.gradle
@@ -56,6 +56,7 @@ dependencies {
 
     testImplementation "junit:junit:$junit_ver"
     testImplementation "org.mockito:mockito-core:$mockito_ver"
+    testImplementation "org.powermock:powermock-mockito-release-full:$powermock_ver"
     testImplementation "com.noveogroup.android:android-logger:$android_logger_ver"
 
     androidTestImplementation "androidx.test.ext:junit:$androidx_test"
--- a/event-handler/build.gradle
+++ b/event-handler/build.gradle
@@ -51,6 +51,8 @@ dependencies {
     api project(':shared')
     implementation "androidx.annotation:annotation:$annotations_ver"
     implementation "androidx.work:work-runtime:$work_runtime"
+    // Base64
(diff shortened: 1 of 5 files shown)


event-handler/build.gradle
```diff
@@ -51,11 +51,14 @@ dependencies {
     api project(':shared')
     implementation "androidx.annotation:annotation:$annotations_ver"
     implementation "androidx.work:work-runtime:$work_runtime"
+    // Base64
+    implementation "org.apache.commons:commons-collections4:4.1"
```
mnoman09, line 55: can we use this library for base64 
```suggestion
    implementation "commons-codec:commons-codec:1.15"
```
As this is the direct dependency we want to use also we are facing dependency issue with currently used library in testapp. 
I did the change and tested it with testapp and it's working fine.

event-handler/src/main/java/com/optimizely/ab/android/event_handler/EventHandlerUtils.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class EventHandlerUtils {
+
+    private static final int BUFFER_SIZE = 32*1024;
+
+    public static String compress(@NonNull String uncompressed) throws IOException {
+        byte[] data = uncompressed.getBytes();
+
+        final Deflater deflater = new Deflater();
+        //deflater.setLevel(BEST_COMPRESSION);
```
The-inside-man, line 31: Should we remove commented out code before moving to Master?

The-inside-man (approved):
LGTM - just one small comment for clean up of commented code

PR #395, opened 2021-12-14
chore: clean up old settings in build.gradle

jaeopt:
## Summary
- clean up logger-related old settings in build.gradle


jaeopt:
This fix causes a build issue with android-testapp (dependency).

PR #397, opened 2021-12-21
fix(DatafileHandler): unschedule background polling worker when interval is not set

jaeopt:
## Summary
- This fixes the bug that a polling worker scheduled before is not disabled when Optimizely is initialized again with polling disabled.
- When Optimizely is initialized, stop the previous worker and then restart it again only when polling is enabled.

## Test plan
- Unit tests

## Issues
- OASIS-8118


mnoman09 (approved):
LGTM

PR #401, opened 2022-01-19
fix(datafileRescheduler): move config file read to a background thread

jaeopt:
## Summary
- Move config-file read to a background thread when DatafileRescheduler is invoked. 
  This will fix a potential source of ANRs.

## Test plan
- Add unit tests validating datafile worker scheduled properly from DatafileRescheduler.

## Issues
- OASIS-8128

### DIFF
--- a/datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileRescheduler.java
+++ b/datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileRescheduler.java
@@ -66,8 +66,6 @@ public class DatafileRescheduler extends BroadcastReceiver {
                     LoggerFactory.getLogger(BackgroundWatchersCache.class));
             Dispatcher dispatcher = new Dispatcher(context, backgroundWatchersCache, LoggerFactory.getLogger(Dispatcher.class));
             dispatcher.dispatch();
-
-
         }  else {
             logger.warn("Received invalid broadcast to data file rescheduler");
         }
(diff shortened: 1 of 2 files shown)


datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileRescheduler.java
```diff
@@ -91,21 +91,35 @@ public void onReceive(Context context, Intent intent) {
...
-                // for scheduled jobs Android O and above, we use the JobScheduler and persistent periodic jobs
-                // so, we don't need to do anything.
-               if (android.os.Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
-                   WorkerScheduler.scheduleService(context,
-                           DatafileWorker.workerId + datafileConfig.getKey(),
-                           DatafileWorker.class,
-                           DatafileWorker.getData(datafileConfig),
-                           DefaultDatafileHandler.getUpdateInterval(context));
-                    logger.info("Rescheduled data file watching for project {}", datafileConfig);
+            Thread thread = new Thread() {
```
msohailhussain, line 94: Can multiple `dispatch` be called? if yes is this the expected behavior while one datafile fetch is in progress and 2nd time dispatch is called and multiple time datafiles are fetched.

msohailhussain (approved):
lgtm. adding @mnoman09 to review this as well.

datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileRescheduler.java
jaeopt: @msohailhussain This can be called for reboot or reinstall, so not likely called repeatedly. Also, reschedule will cancel existing one before creating a new task, so it must be safe.

PR #408, opened 2022-02-09
chore: remove logger properties from sdk

jaeopt:
## Summary
- Remove android-logger.properties file (noveogroup) from sdk. App is supposed to have own logger properties file.


jaeopt:
It looks like we better keep the redundant logger property file as a default config.

PR #410, opened 2022-03-15
fix: set client-engine version in events

jaeopt:
## Summary
- OptimizelyManager overrides the client name/version (in events) with [the updated API of java-sdk core](https://github.com/optimizely/java-sdk/pull/466). The client version was not overridden properly in the previous versions.

## Test plan
- A new test validating client name/version in events.

## Issues
- OASIS-8188


build.gradle
```diff
@@ -19,13 +19,11 @@
 buildscript {
     ext.kotlin_version = '1.4.0'
     
-    def version_name = System.getenv('TRAVIS_TAG')
-    if (version_name != null) {
-        rootProject.ext.version_name = version_name
-    } else {
-        rootProject.ext.version_name = 'debugVersion'
+    ext.version_name = System.getenv('TRAVIS_TAG')
```
msohailhussain, line 22: what is it for?

msohailhussain (approved):
lgtm

build.gradle
jaeopt: @msohailhussain We get the sdk release version from a git tag forwarded by travis (TRAVIS_TAG). This change is to make it work for testing in travis as well (which fails to override with test version).

PR #400, opened 2022-01-18
chore: fix to support android-12+

jaeopt:
## Summary
Fix for android 12+ support (app crashes with PendingIntent flag requirement when built with targetSdkVersion=31)
- upgrade dependency "androidx.work:work-runtime" to 2.7.1
- add "android:exported" to test-app activities
- upgrade compileSDKVersion/targetSDKVersion to 32

## Test plan
- Build/run test-app with Android OS 31 and 32

## Issues
- OASIS-8137

PR #411, opened 2022-03-17
chore: removed travis yml and added git action support

mnoman09:
## Summary
- We remove travis yaml from android and moved all the stages to git action.

## Test plan
All tests should pass


.github/workflows/android.yml
```diff
@@ -0,0 +1,180 @@
...
+    - name: Clean all modules
+      run: ./gradlew cleanAllModules
+    - name: Build with Gradle
+      run: ./gradlew build
+  test:
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        api-level: [21, 25, 26, 29]
```
jaeopt, line 92: I understand that github has an issue supporting AVD 31+. 
Can we change "compile_sdk_version" to 31 (no need for "target_sdk_version") since it's required to run other fix?

.github/workflows/android.yml
```diff
@@ -0,0 +1,180 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'temurin'
+        cache: gradle
```
jaeopt, line 80: These steps are repeated for all tasks. Just wondering if we can factor out as a function to simplify this workflow file. github needs to improve it :)

jaeopt (changes requested):
Looks great! I have a couple of suggestions/clarification.
Also please fix the PR title (it should be "chore: remove...") and description (empty)

.github/workflows/android.yml
mnoman09: Yes I think we can change android compile sdk version to 31 it should cause no problem. As we already tested that on PR #406

.github/workflows/android.yml
mnoman09: Sure we can refactor it and move to seperate yaml then use it as uses. As of right now I don't there is any way to reuse the steps by keeping it in same file.

mnoman09: I think with current implementation it is difficult to add reusable workflow so can we refactor it later if possible?

jaeopt: Sounds good to me!

mnoman09:
> @mnoman09 I see unit tests fail. Can we try to revert the last commit and then upgrade compile_sdk_version only (keeping target_sdk_version to 29)? The fix to support target_sdk_version 31+ will be handled in a separate PR (#400).

I made the changes as requested.

jaeopt (approved):
LGTM

PR #412, opened 2022-03-18
chore: upgrade mockito and tests for jdk11 cleanup

jaeopt:
## Summary
- Upgrade mockito (3.12) and powermock(2.0) to clean up JDK11 warnings
- Fix tests accordingly


jaeopt:
mockito 3.x requires android API 24+. It can be used for the sdk with minSdkVersion = 14. 
We'll hold this upgrading until we boost the minSdkVersion to 24 or higher.

PR #415, opened 2022-04-01
fix: change default attributes to more meaningful values

jaeopt:
## Summary
- (android SDK package name) -> android SDK version 
- (app version + serail-id) -> app version

These default attributes are filtered out from events unless attributes are set in the project configuration.


msohailhussain (approved):
lgtm

Issue #419, opened 2022-04-19
StrictMode violation due to Deflater.end() not being called


schmecs:
Minor / no obvious effect on performance. `EventHandlerUtils` isn't calling `end()` on its `deflater` when finished which causes a StrictMode violation.

See logs:

```
D/StrictMode: StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
        at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1987)
        at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:345)
        at java.util.zip.Deflater.finalize(Deflater.java:575)
        at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:291)
        at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:278)
        at java.lang.Daemons$Daemon.run(Daemons.java:139)
        at java.lang.Thread.run(Thread.java:920)
     Caused by: java.lang.Throwable: Explicit termination method 'end' not called
        at dalvik.system.CloseGuard.openWithCallSite(CloseGuard.java:295)
        at dalvik.system.CloseGuard.open(CloseGuard.java:263)
        at java.util.zip.Deflater.<init>(Deflater.java:189)
        at java.util.zip.Deflater.<init>(Deflater.java:206)
        at com.optimizely.ab.android.event_handler.EventHandlerUtils.compress(EventHandlerUtils.java:30)
        at com.optimizely.ab.android.event_handler.EventWorker.compressEvent(EventWorker.java:114)
        at com.optimizely.ab.android.event_handler.EventWorker.getData(EventWorker.java:94)
        at com.optimizely.ab.android.event_handler.EventWorker.getData(EventWorker.java:100)
        at com.optimizely.ab.android.event_handler.DefaultEventHandler.dispatchEvent(DefaultEventHandler.java:99)
        at com.optimizely.ab.event.BatchEventProcessor$EventConsumer.flush(BatchEventProcessor.java:226)
        at com.optimizely.ab.event.BatchEventProcessor$EventConsumer.addToBatch(BatchEventProcessor.java:189)
        at com.optimizely.ab.event.BatchEventProcessor$EventConsumer.run(BatchEventProcessor.java:164)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:462)
[...]

PR #422, opened 2022-04-28
fix: call end inflater/deflater to avoid leaks

jaeopt:
## Summary
- StrictMode detects leaks with inflater and deflater.
- Add end() calls at the end of inflate/deflate

## Test plan
- Pass with existing compress/decompress
- Test with StrictMode 

## Issues
- [ISSUE-419](https://github.com/optimizely/android-sdk/issues/419)


mikechu-optimizely (commented):
The `finally` block additions make sense to me. (comment only)

PR #431, opened 2022-09-12
feat(ATS): add vuid manager

jaeopt:
## Summary
Add VuidManager which will create and manage a vuid:
 
- create a random vuid if it cannot be found in the device with the required format.
- "vuid_" + 32 hexadecimal string (UUID with all "-"s removed). 
- all lowercases ("vuid_330e05cad15746d9af8a75b8d1050a87")
- SDK will reuse the same vuid (saving to and retrieving from a persistent storage when the client application starts).
- SDK will support a single vuid shared by multiple SDK instances (thread-safe)
- ODP module added

## Test plan
- add unit tests

## Issues
- [FSSDK-8548](https://optimizely.atlassian.net/browse/FSSDK-8548)


odp/src/androidTest/java/com/optimizely/ab/android/odp/VuidManagerTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        // remove vuid storage
+        cleanSheredPrefs()
+        // remove a singlton instance
+        VuidManager.removeSharedForTesting()
+
+        vuidManager = VuidManager.getShared(context)
+    }
+
+    @After
+    fun cleanSheredPrefs() {
```
opti-jnguyen, line 43: Should this be `cleanSharedPrefs`?
Typo in "Shered" to "Shared".

event-handler/src/androidTest/java/com/optimizely/ab/android/event_handler/EventDAOTest.java
```diff
@@ -54,11 +54,12 @@ public void setupEventDAO() {
         logger = mock(Logger.class);
         context = InstrumentationRegistry.getInstrumentation().getTargetContext();
         eventDAO = EventDAO.getInstance(context, "1", logger);
+        tearDownEventDAO();
     }
 
     @After
     public void tearDownEventDAO() {
-        assertTrue(context.deleteDatabase(String.format(EventSQLiteOpenHelper.DB_NAME , "1")));
+        context.deleteDatabase(String.format(EventSQLiteOpenHelper.DB_NAME , "1"));
```
opti-jnguyen, line 62: Out of curiosity, why was the assertion removed here?

test-app/src/main/java/com/optimizely/ab/android/test_app/NotificationService.kt
```diff
@@ -20,7 +20,8 @@ import android.content.Intent
 import androidx.core.app.NotificationCompat
 import com.optimizely.ab.android.shared.CountingIdlingResourceManager
 
-class NotificationService : IntentService("NotificationService") {
+class
```
opti-jnguyen, line 23: Is `class` intentionally on a separate line here?

opti-jnguyen (changes requested):
Overall LGTM besides the probable typo with `shered`.

Food for thought - what do you think about abstracting out VUID as its own class, separate from VUIDManager?

Was fun re-visiting Kotlin!

event-handler/src/androidTest/java/com/optimizely/ab/android/event_handler/EventDAOTest.java
jaeopt: For cleaning up with remaining database from other tests, the teardown is also called when starting each test. So assert is not true always.

test-app/src/main/java/com/optimizely/ab/android/test_app/NotificationService.kt
jaeopt: Not intentional. restored :)

opti-jnguyen (approved):
LGTM

Issue #429, opened 2022-08-30
DatafileClient should specify read/write timeout on urlConnection


pdudchenok:
Recently we observed significant amount of OOM  at our application due to the following cause 
>java.lang.OutOfMemoryErrorThread.java:730Could not allocate JNI Env

The issue is happening due to limitation on amount of threads per process.

We have noticed that a lot of threads are stuck at `SSL_do_handshake`.
While inspecting the source code we observed that 
`urlConnection` created [here](https://github.com/optimizely/android-sdk/blob/2487c3e67258fc9651ee40ed72a3e56aaaf5be4c/datafile-handler/src/main/java/com/optimizely/ab/android/datafile_handler/DatafileClient.java#L77) doesn't have  configured read/write socket timeout. By [default](https://developer.android.com/reference/java/net/URLConnection#getReadTimeout()) the values are ` 0` which means infinity. 

The full trace:
[...]

PR #432, opened 2022-09-16
fix(DatafileClient): add timeouts to UrlConnections

jaeopt:
## Summary
Set timeouts to UrlConnections for DatafileClient and EventClient so they can release resources when connection failed (0 by default, which is no timeout).

## Test plan
- Add unit tests for UrlConnections timeout

## Issues
- [OASIS-8568](https://jira.sso.episerver.net/browse/FSSDK-8568)
- https://github.com/optimizely/android-sdk/issues/429


msohailhussain (approved):
lgtm

PR #434, opened 2022-10-21
fix: clean up old codes for JobScheduler and Services.

jaeopt:
## Summary
Since all background tasks have been changed to WorkManager, we do not need old codes for JobScheduler, DatafileService and EventService. All these codes and related tests are cleaned up.

## Test plan
- All existing unit and FSC tests should pass.

## Issues
- FSSDK-8637


mnoman09 (approved):
LGTM

PR #438, opened 2022-11-21
Fix: Reverted putting namespace in gradle file instead of manifiest 

mnoman09:
## Summary
- Reverted putting namespace in gradle file instead of manifiest (see PR 437)

## Issues
- FSSDK-8688


jaeopt (approved):
LGTM

PR #440, opened 2022-11-29
feat(ATS): add ODP API clients

jaeopt:
## Summary
Add ODP API clients for segment fetch and event dispatch:
- DefaultODPAPIManager
- ODPSegmentClient
- ODPEventClient

## Test plan
- Unit tests

## Issues
- FSSDK-8546, FSSDK-8547


mnoman09 (approved):
This implementation looks good. Just have one question that is the final and complete implementation of ODP in android SDK? Or will there be a separate PR in which, we will expose the sendODPEvent and registration of ODP manager functions?

jaeopt:
> This implementation looks good. Just have one question that is the final and complete implementation of ODP in android SDK? Or will there be a separate PR in which, we will expose the sendODPEvent and registration of ODP manager functions?
@mnoman09 there will be one more PR for top-level integration as you expect.

Issue #458, opened 2023-05-25
ResourcesNotFoundException thrown during initialisaiton


barchynskyi:
SDK version: 3.10.1
Android version: multiple

Recently we have detected issue during the SDK init. Looks like Optimizely throws exception outside error handler and leads to broken and unrecoverable Optimizely state. We could not identify steps to reproduce the issue but it looks like it happens during synchronous initialisation.

Stacktrace:
`android.content.res.Resources$NotFoundException: Resource ID #0xffffffff
                                                                                                    	at android.content.res.ResourcesImpl.getValue(ResourcesImpl.java:269)
                                                                                                    	at android.content.res.ResourcesImpl.openRawResource(ResourcesImpl.java:425)
                                                                                                    	at android.content.res.Resources.openRawResource(Resources.java:1461)
                                                                                                    	at android.content.res.Resources.openRawResource(Resources.java:1405)
                                                                                                    	at com.optimizely.ab.android.sdk.OptimizelyManager.loadRawResource(OptimizelyManager.java:457)
                                                                                                    	at com.optimizely.ab.android.sdk.OptimizelyManager.safeLoadResource(OptimizelyManager.java:387)
                                                                                                    	at com.optimizely.ab.android.sdk.OptimizelyManager.getDatafile(OptimizelyManager.java:340)
                                                                                                    	at com.optimizely.ab.android.sdk.OptimizelyManager.initialize(OptimizelyManager.java:264)`

Tamara-Barum:
Opening an internal ticket to review:
FSSDK-9406

mnoman09:
@barchynskyi could you please share the code which you are using for OptimizelySDK initialization?

Tamara-Barum:
Issue has been confirmed resolved..

PR #460, opened 2023-07-07
[FSSDK-9406] fix: Added catch block to capture resource not found exception

mnoman09:
## Summary
- This will gracefully log the exception that was reported in issue #458 

## Test plan
All tests should pass
## Issues
- [FSSDK-9406](https://jira.sso.episerver.net/browse/FSSDK-9406)


jaeopt (approved):
LGTM

PR #471, opened 2024-01-08
[FSSDK-9917] update proguard rule for ODP

jaeopt:
## Summary

- Release build with R8 (AGP 8) fails with missing java.beans.Transient. SDK consumer proguard rules are updated to support it.
- Upgrade Java source and target version from 8 to 11.

## Issues
- [FSSDK-9917](https://jira.sso.episerver.net/browse/FSSDK-9917)


andrewleap-optimizely (approved):
LGTM

PR #483, opened 2024-07-30
[FSSDK-10501] forward exceptions for ODP fetch segments

jaeopt:
## Summary
- Forward exceptions for ODP fetch segments to support debugging in the application.

## Issues
- "FSSDK-10501"


Mat001 (approved):
LGTM!

PR #491, opened 2024-08-22
[FSSDK-10587] chore: rollback to 4.0.0

muzahidul-opti:
## Summary
- Rollback to 4.0.0

The "why", or other context.

## Issues
- [FSSDK-10587](https://jira.sso.episerver.net/browse/FSSDK-10587)


jaeopt (changes requested):
It looks like reverting 4.0.2 only, not 4.0.1.

muzahidul-opti:
> It looks like reverting 4.0.2 only, not 4.0.1.

Oh, I thought only forwarding exception part should be updated. 
So, we need to remove `ClientForODPOnly` class also, right?

jaeopt (approved):
LGTM! we can fix FSSDK-000 and confirm unit tests passed

PR #495, opened 2024-09-23
[FSSDK-10665] fix: Github Actions YAML files vulnerable to script injection

FarhanAnjum-opti:
## Summary
Instead of directly using variables that can contain user input, new intermediate variables were introduced so that script injection attacks could be mitigated. Also removed unnecessary use of ${{ }} as using it makes the file vulnerable.

## Issues
- [FSSDK-10665](https://jira.sso.episerver.net/browse/FSSDK-10665)


muzahidul-opti (approved):
LGTM

PR #497, opened 2024-10-31
 [FSSDK-10758] implement vuid configuration independent of odp

raju-opti:
## Summary
- With this PR, VUID can be enabled/disabled independently of ODP

## Test plan
- added news tests and updated affected old tests

## Issues
- FSSDK-10758

### DIFF
--- a/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
+++ b/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java
@@ -80,7 +80,7 @@ public class OptimizelyClient {
         So, we start with an empty map of default attributes until the manager is initialized.
         */
 
-        if (isValid()) {
+        if (isValid() && vuid != null) {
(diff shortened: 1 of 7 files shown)


muzahidul-opti (commented):
Looks good to me, where do we check client_initialized event based on vuid enable status.

raju-opti:
> Looks good to me, where do we check client_initialized event based on vuid enable status.

this was already implemented here: https://github.com/optimizely/android-sdk/blob/eebed3cbd31400c98e9309ec6c91e131e3be5b0e/android-sdk/src/main/java/com/optimizely/ab/android/sdk/OptimizelyClient.java#L83

This will drop any event if there is no valid identifier but there will be a error log. I have added a new condition to check whether the vuid is not null to avoid the log: https://github.com/optimizely/android-sdk/pull/497/commits/3512a6582877f7a443698b15f41ace8fd1af3e1b

@muzahidul-opti

jaeopt (approved):
LGTM - with a small fix suggestion

PR #512, opened 2025-08-20
[FSSDK-11539] feat: holdout feature support added

muzahidul-opti:
## Summary
- Core functionality is implemented by java-sdk.
- Add test cases for holdout support.
- Unit test is failing because holdout feature didn't publish yet for java.

## Note
The PR would be merged into master after publishing java holdout feature.

## Issues
- FSSDK-11539

### DIFF
--- a/android-sdk/build.gradle
+++ b/android-sdk/build.gradle
@@ -59,6 +59,13 @@ android {
     }
 }
 
+repositories {
+    maven {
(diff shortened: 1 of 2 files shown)


muzahidul-opti:
> Looks good, but not 100% sure if we need this test in Android (and flutter) level. Unless we have any reason not working in android, we can drop?

Good point. I think so, we may drop the test cases.

PR #513, opened 2025-08-27
[FSSDK-11782] chore: migrate publishing process from OSSRH to maven central portal

muzahidul-opti:
## Summary
- OSSRH has reached end of life on June 30, 2025.
- To minimize changes while enabling compatibility with the new portal, add the io.github.gradle-nexus.publish-plugin.          - This plugin automates staging management and works with the OSSRH Staging API for transitional support.
- https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#overview

## Issues
- FSSDK-11782

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,8 +26,8 @@ buildscript {
     ext.is_release_version = !version_name.endsWith("SNAPSHOT")
 
     repositories {
-        jcenter()
         google()
+        mavenCentral()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:7.2.2'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -117,9 +117,35 @@ task testODPModule () {
...
+            sonatypeTasks.add('closeSonatypeStagingRepository')
+        }
+
+        def args = ['./gradlew']
+        args.addAll(sonatypeTasks)
+        args.add('--info')
+
+        exec {
+            commandLine args
+        }
```
jaeopt, line 147: why do we have a different step here for android only?  I do not see this step for java.

jaeopt (changes requested):
Need a clarification. See my comments.

build.gradle
muzahidul-opti: Ideally for the snapshot version we don't need `closeSonatypeStagingRepository`, it will directly go to snapshot repository, not staging. 
For java I didn't separate it, because it doesn't cause any issue, but for Android it doesn't work. I think due to multiple modules publishing cause timing issue.

jaeopt (approved):
LGTM

PR #536, opened 2026-07-01
[AI-FSSDK] [FSSDK-12750] Use attribute id instead of key for CMAB predictions

raju-opti:
## Summary

The CMAB prediction request payload requires the attribute ID in the `id` field, but the attribute key was being sent instead. The Android SDK inherits CMAB logic from the Java SDK core-api, which already contains this fix. This PR bumps the Java core dependency version and updates tests to validate attribute ID usage.

## Changes
- Bumped Java SDK core dependency to pick up the attribute ID fix from core-api
- Updated CMAB client test attributes to use ID-based keys instead of key-based keys
- Added test verifying the JSON request payload uses attribute IDs

## Jira Ticket
[FSSDK-12750](https://optimizely-ext.atlassian.net/browse/FSSDK-12750)

[FSSDK-12750]: https://optimizely-ext.atlassian.net/browse/FSSDK-12750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android-sdk/src/test/java/com/optimizely/ab/android/sdk/cmab/DefaultCmabClientTest.java
+++ b/android-sdk/src/test/java/com/optimizely/ab/android/sdk/cmab/DefaultCmabClientTest.java
@@ -1,4 +1,4 @@
-// Copyright 2025, Optimizely, Inc. and contributors
+// Copyright 2025, 2026, Optimizely, Inc. and contributors
 //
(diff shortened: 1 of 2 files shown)


muzahidul-opti (approved):
LGTM

